PR #35, opened 2017-08-08
Improve PackageMetadataMVC tests

ilayaperumalg:
- Add an explicit skipperHome property value for the tests so that the default isn't used
 - Remove deleteAll before tests
 - Remove Transactional and test method ordering


markpollack:
Fixed via 523445b27747a2301ec68e8171fa6ac67ed04f24

PR #39, opened 2017-08-10
Use Spring Shell 2 instead of Shell 1.

ericbottard:
Rewrote the two exiting commands, as well as some supporting classes.

ALSO:
* Fix version numbers (used a dash instead of a dot)
* Remove usage of versions of modules, by gathering everything in parent dependencyMgt
* Remove Boot JPA starter in favor of just JPA API jar


ilayaperumalg:
@ericbottard For some reason, the shell jar exits as soon as it shows up the banner. Could you check if you see the same?

ericbottard:
Hmm, that's weird, getting this too now, even if I used to run with `mvn spring-boot:run`.

Seems to be about tomcat not being able to start, which points to the fact that the shell should not even embed tomcat. Comes from `spring-cloud-skipper` that has web support but shouldn't.

Investigating, I'll update the PR and let you know.

ericbottard:
@ilayaperumalg Pushed a fix

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as e75b168c0f521534f6b7fe128b60026cec8019ff

PR #74, opened 2017-09-14
Update operation changes

### LINKED ISSUE #72: Update operation should use UpdateProperties instead of DeployProperties
Currently, the update operation uses DeployProperties which is used by the `deploy` operation. We need to have a separate properties `UpdateProperties` so that we could have more control over update related properties (dry-run etc., in future)

### LINKED ISSUE #73: Move update operation into Release controller
Currently, the `update` operation is done via PackageController. Given the `update` operation is done on a release, it makes sense to move this operation into `ReleaseController`.

ilayaperumalg:
- Move update to ReleaseController
 - Use UpdateProperties instead of DeployProperties
 - Use UpdateRequest that wraps PackageIdentifier and UpdateProperties
 - Update ReleaseService to handle UpdateRequest
 - Update tests

Resolves #72
Resolves #73


ilayaperumalg:
Rebased.

markpollack:
Made some additional improvements, squashed and merged as 9a6e04cc40c5034c3e2953035ae3b5756baa4772

PR #97, opened 2017-09-20
Do not use packageMetadata.getOrigin as path when loading packages

### LINKED ISSUE #96: Do not use packageMetadata.getOrigin as path when loading packages
Use the path right after unzipping the package.
Change downloadPackage to return the Package.

markpollack:
* Remove PackageService.loadPackage method, downloadPackage now returns Package
* application.yml for tests specify their own local skipperHome to use local 'target' directory
* Default application.yml specifies spring snapshot repo for maven to simplify development.
* Add ignoring of "origin" in package metadata, going to be implemented correctly in another issue shortly
* Move @ConfigurationProperties for Maven to a top level class
* Do not directly instantiate MavenConfigurationProperties

Fixes #96


ilayaperumalg:
LGTM, merged as d4b04f56a3279b224a62aad2133d5b8223152cc0

PR #109, opened 2017-09-26
Switch from in memory key-value store to JPA

### LINKED ISSUE #106: Replace map based repositories with JPA
`PackageMetadata` and `Release` use spring data map backed repo.  Look for @KeySpace and replace with JPA annotations.

markpollack:
* PackageMetadata and Release are now @Entity along with nested classes
* Add AbstractEntity base class
* In Release entity, Serialize Package, ConfigValues, and Manifest as Lob
* Keep AppDeployerData as in-memory since it is used internally and
  initialized when the application starts.  We don't want to store user creds
  in the DB.

Fixes #106

### DIFF
--- a/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/PackageMetadata.java
+++ b/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/PackageMetadata.java
@@ -15,23 +15,17 @@
  */
 package org.springframework.cloud.skipper.domain;
 
+import javax.persistence.Entity;
 import javax.persistence.Lob;
 import javax.validation.constraints.NotNull;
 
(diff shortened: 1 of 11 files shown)


spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/PackageMetadata.java
```diff
@@ -151,13 +145,13 @@ public void setVersion(String version) {
...
 	}
 
-	public String getId() {
-		return id;
-	}
-
-	public void setId(String id) {
-		this.id = id;
-	}
+	// public String getId() {
```
markpollack: These commented out line should be deleted

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as 288ce9c

PR #111, opened 2017-09-26
Support config values override

### LINKED ISSUE #103: Install/update commands to take --file and --set to specify properties to override in the package
Allows runtime overrides.  Have not tested end to end.

Tests are in ConfigValueUtilsTests.  ATM, the test does not take into account 'top level' overrides, 

e.g.
```
Package pkg = packageService.loadPackageOnPath(resource.getFile());
Map<String, Object> mergedMap = ConfigValueUtils.mergeConfigValues(pkg, new ConfigValues());
```
Instead of an empty ConfigValues, some data should be put in the ConfigValues 'raw' field and test to see it takes precedence.

example usage in shell, which needs to be added.
```
package deploy --name log --version 2.0.0 --release-name mylog --set appVersion=1.4.0.RELEASE

log-2.0.0 has

appVersion: 1.2.0.RELEASE

in values.yml

The above command would override that values and set appVersion to be 1.4.0.RELEASE

specify the same package version, but change the appVersion

package update --release-name mylog --package-name log --package-version 2.0.0 --set appVersion=1.5.0.RELEASE
```

POC version of skipper had a propertiesToYaml utility method based on this library

https://github.com/Codearte/props2yaml

See the helm https://github.com/kubernetes/helm/blob/master/docs/using_helm.md for example usage of --set and --values (aka --file)

ilayaperumalg:
- install/upgrade commands now support setting config values set via `--file` as YAML file or via `--set` with comma separated String properties from the shell

Resolves #103

### DIFF
--- a/spring-cloud-skipper-shell-commands/pom.xml
+++ b/spring-cloud-skipper-shell-commands/pom.xml
@@ -44,6 +44,16 @@
             <version>2.5</version>
         </dependency>
         <dependency>
+            <groupId>io.codearte.props2yaml</groupId>
+            <artifactId>props2yaml</artifactId>
+            <version>0.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>1.18</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
This is merged as 784cbdc

PR #133, opened 2017-10-05
Add release status updated

### LINKED ISSUE #122: Add background thread that polls deployed releases for status
The status returned when querying for releases should reflect the current state on the platform.  We can do this by polling and storing the most recent state.  

Since apps that were deployed recently are most likely to be queries for quickly to see if they are running, the polling frequency for apps deployed < 2-3 minutes ago should be higher.

The status command can still get compute the state on demand, as it does now.  Perhaps it should then also update the state in the database to avoid an inconsistency when viewing state via other commands.

jvalkeal:
- As working with different release version, add
  optional version property to status command.
- New ReleaseStateUpdateService which polls
  existing apps and reflects status back to
  ReleaseRepository.
- Tries to update new apps more frequently and
  doing full poll less frequently.
- Fixes #122

### DIFF
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
@@ -117,10 +117,18 @@ public class DefaultSkipperClient implements SkipperClient {
 	}
 
 	@Override
-	public Info status(String releaseName) {
+	public Info status(String releaseName, Integer releaseVersion) {
 		Map<String, String> uriVariables = new HashMap<String, String>();
 		uriVariables.put("releaseName", releaseName);
-		return this.restTemplate.getForObject(baseUri + "/status/{releaseName}", Info.class, uriVariables);
+		if (releaseVersion == null) {
+			return this.restTemplate.getForObject(baseUri + "/status/{releaseName}", Info.class, uriVariables);
+		}
+		else {
+			uriVariables.put("releaseVersion", Integer.toString(releaseVersion));
+			Release release = this.restTemplate.getForObject(baseUri + "/status/{releaseName}/{releaseVersion}",
+					Release.class, uriVariables);
+			return release.getInfo();
+		}
 	}
 
 	@Override
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
@@ -117,18 +117,20 @@ public class DefaultSkipperClient implements SkipperClient {
 	}
 
 	@Override
-	public Info status(String releaseName, Integer releaseVersion) {
+	public Info status(String releaseName) {
 		Map<String, String> uriVariables = new HashMap<String, String>();
 		uriVariables.put("releaseName", releaseName);
-		if (releaseVersion == null) {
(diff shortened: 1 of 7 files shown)


spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
```diff
@@ -117,10 +117,18 @@ public AboutInfo info() {
 	}
 
 	@Override
-	public Info status(String releaseName) {
+	public Info status(String releaseName, Integer releaseVersion) {
```
markpollack: Should we add an overloaded method instead?  That is provide
`status(String releaseName)` as well as `status(String releaseName, Integer releaseVersion)` in which case we would assert not null for releaseVersion?

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/deployer/AppDeployerReleaseManager.java
```diff
@@ -107,34 +112,67 @@ public Release install(Release releaseInput) {
...
+			if (deployedCount == deploymentIds.size()) {
 				release.getInfo().getStatus().setPlatformStatus("All the applications are deployed successfully.");
 			}
+			else if (unknownCount == deploymentIds.size()) {
+				release.getInfo().getStatus().setPlatformStatus("All applications unknown to system.");
+			}
 			else {
-				release.getInfo().getStatus().setPlatformStatus(
-						"Applications deploying... " + releaseStatusMsg.toString());
+				release.getInfo().getStatus().setPlatformStatus("State of applications: "
```
markpollack: We should give some indication that there maybe apps that are in the state deployed and unknown.  In the work for 'smart update/rollback', I've changed what is stored `appDeployerData.getDeploymentData()` to be a Map of appName, deploymentId which makes it clear what application name is in what state.  In the meantime, maybe we can gather up `appInstanceStatus.getId()` for all the deployed apps and show it in the case of 'some apps deployed, some apps still deploying`.

markpollack:
Made small tweaks to formatting so it fits better within a fixed width when there are multiple apps
Squashed and merged as f3ec7efe41864f3025241abcb4cd80ca5ef6a854
```
skipper:>status --release-name mylog3
╔═══════════════╤════════════════════════════════════════╗
║Last Deployed  │Fri Oct 06 12:16:26 EDT 2017            ║
║Status         │DEPLOYED                                ║
║Platform Status│State of applications:                  ║
║               │mylog3-v1.log=[mylog3-v1.log-1=deploying║
║               │mylog3-v1.log-0=deploying]              ║
╚═══════════════╧════════════════════════════════════════╝

skipper:>status --release-name mylog3
╔═══════════════╤════════════════════════════════════════╗
║Last Deployed  │Fri Oct 06 12:16:26 EDT 2017            ║
║Status         │DEPLOYED                                ║
║Platform Status│State of applications:                  ║
║               │mylog3-v1.log=[mylog3-v1.log-1=deploying║
║               │mylog3-v1.log-0=deployed]               ║
╚═══════════════╧════════════════════════════════════════╝

skipper:>status --release-name mylog3
╔═══════════════╤═══════════════════════════════════════════════╗
║Last Deployed  │Fri Oct 06 12:16:26 EDT 2017                   ║
║Status         │DEPLOYED                                       ║
║Platform Status│All the applications are deployed successfully.║
║               │mylog3-v1.log=[mylog3-v1.log-1=deployed        ║
║               │mylog3-v1.log-0=deployed]                      ║
╚═══════════════╧═══════════════════════════════════════════════╝
```

PR #136, opened 2017-10-06
Support for 'smart updates'

### LINKED ISSUE #105: "Smart" application version update
Determine the diff between two releases when upgrading and only take the minimal actions necessary.

markpollack:
Support updating only the parts of a multiple dependency package that have changed.

* Add `ReleaseAnalysisService` to determine changes to make on update
* Create `UpdateStrategy` to provide multiple ways to update the application
* Introduced guava as a dependency (ducking)

Fixes #105

Tested watching `ReleaseAnalysisServiceTests` execute and ad-hoc via starting server
```
java -jar spring-cloud-skipper-server/target/spring-cloud-skipper-server-1.0.0.BUILD-SNAPSHOT.jar --spring.config.location=/home/mpollack/skipper.yml
```
and the following shell commands
```
skipper:>install --name ticktock --version 1.0.0 --release-name mytick
Released mytick
skipper:>upgrade --release-name mytick --package-name ticktock --package-version 1.0.0  --file /home/mpollack/log-update.yml
mytick has been upgraded.
Last Deployed: Thu Oct 05 22:29:35 EDT 2017
Status: Applications deploying... mytick-v2.log=[mytick-v2.log-0=deploying]
```
Where `log-update.yml` can be
```
log:
  version: 1.2.0.RELEASE
```
and `jps` will show the updated version of the log process, while the time stays the same.

Alternatively `log-update.yml` can be
```
log:
  spec:
    applicationProperties
      log.level: ERROR
```
And you can access the (secured) boot `env` endpoint for the log application and see the value `ERROR`.

Note: in this ticktock example, since no stream binding properties were specified, there is no tick or tocking.

An inline `--set` command that specified yaml in a properties style format will be added later.


spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/deployer/ReleaseAnalysisReport.java
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.skipper.deployer;
+
+import java.util.List;
+
+import io.jsonwebtoken.lang.Assert;
```
markpollack: wrong intellij assert import.

spring-cloud-skipper-server/src/test/java/org/springframework/cloud/skipper/service/NoOpAppDeployer.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.cloud.deployer.spi.app.AppStatus;
+import org.springframework.cloud.deployer.spi.app.DeploymentState;
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
+import org.springframework.cloud.deployer.spi.util.RuntimeVersionUtils;
+
+/**
+ * @author Mark Pollack
+ */
+public class NoOpAppDeployer implements AppDeployer {
```
markpollack: Not used now, maybe delete.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/deployer/strategies/SimpleRedBlackUpgradeStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+import org.springframework.cloud.skipper.domain.Release;
+import org.springframework.cloud.skipper.domain.SpringBootAppKind;
+import org.springframework.cloud.skipper.domain.SpringBootAppKindReader;
+import org.springframework.cloud.skipper.domain.Status;
+import org.springframework.cloud.skipper.domain.StatusCode;
+import org.springframework.cloud.skipper.repository.DeployerRepository;
+import org.springframework.cloud.skipper.repository.ReleaseRepository;
+
+/**
+ * @author Mark Pollack
```
markpollack: Add javadoc

spring-cloud-skipper-server/pom.xml
```diff
@@ -139,6 +133,11 @@
             <artifactId>spring-cloud-deployer-kubernetes</artifactId>
             <version>${spring-cloud-deployer-kubernetes.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
```
jvalkeal: Uh oh, please lord save us from guava! We need to find something else.

spring-cloud-skipper-server/src/test/java/org/springframework/cloud/skipper/AbstractIntegrationTest.java
```diff
@@ -33,4 +39,28 @@
...
+		try {
+			Thread.sleep(5000);
+			for (Release release : releaseRepository.findAll()) {
+				if (release.getInfo().getStatus().getStatusCode() != StatusCode.DELETED) {
+					releaseService.delete(release.getName());
+				}
+			}
+		}
+		catch (Exception e) {
+			System.out.println("error cleaning up resource in integration test");
```
jvalkeal: Even for tests, proper logging is better.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/deployer/strategies/UpgradeStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+	/**
+	 * Given the two releases, the one currently deployed, the 'existingRelease' and to one to
+	 * be deployed, the 'replacingRelace', use the information in the analysis report to
+	 * deploy the apps in the new release and delete the apps in the old.
+	 * @param existingRelease the apps currently deployed
+	 * @param replacingRelease the apps to be deployed
+	 * @param releaseAnalysisReport report to guide the strategy on what apps to replace.
+	 * @return the replacingRelease, now deployed.
+	 */
+	Release upgrade(Release existingRelease, Release replacingRelease, ReleaseAnalysisReport releaseAnalysisReport);
```
jvalkeal: `ReleaseAnalysisReport` is probably pointless in this interface method as strategy can choose its own ways to get from existing release to replacing release.  Actual strategy impl can probably create `ReleaseAnalysisReport` if it needs it?

spring-cloud-skipper-server/pom.xml
jvalkeal: Maybe we could build something from https://github.com/jgrapht/jgrapht. We're probably going to need more graph structures as that is what a complex deployment will look like.

markpollack: tinkerpop also - http://tinkerpop.apache.org/.  I think I'll be moving the `MapDifference`'s of each app into the `ReleaseDifference` instead of storing just a pretty-print of the differences, so a bit more structured.  The graph approach is prob best in the abstract, the right tool, "just" requires learning those tools....  ATM the shapes of the apps we need to support are very simple, deriving from what we want to do with SCDF, but can get more complex if/when skipper expands outside that use-case.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/deployer/strategies/UpgradeStrategy.java
markpollack: Yea, the design is evolving so it can go in a few directions ATM.  I kept it separate with the  idea that the responsibility of the UpdateStrategy would be to "update given what it is told to do" vs. it also figuring it out.  There may also be use-cases where we need to further modify the manifest (not just the yml merging) such as copying application values from the existingRelease to the replacingRelease, and that can happen based on the results of the analysis (hence wanting to keep the MapDifferences in the report).

spring-cloud-skipper-server/src/test/java/org/springframework/cloud/skipper/AbstractIntegrationTest.java
markpollack: righto!

markpollack:
Squashed and merged as b3ecf4aa7af285e916f5d692a3aa19cd19d766e7 with some minor cleanup.

PR #155, opened 2017-10-09
Move base path into /api

### LINKED ISSUE #130: Move controller endpoint under "api"

ilayaperumalg:
- Add `spring.data.rest.base-path=/api` entry into application.yml

Resolves #130

### DIFF
--- a/spring-cloud-skipper-server/src/main/resources/application.yml
+++ b/spring-cloud-skipper-server/src/main/resources/application.yml
@@ -7,6 +7,9 @@ server:
 spring:
   main:
     banner-mode: "off"
+  data:
+    rest:
+      base-path: /api
 
 # TODO - remove, pre M1 development simplicity
 maven:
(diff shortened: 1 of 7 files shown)


spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/controller/SkipperController.java
```diff
@@ -47,7 +47,7 @@
  * @author Ilayaperumal Gopinathan
  */
 @RestController
-@RequestMapping("/")
+@RequestMapping("/api")
```
ilayaperumalg: Ideally, we can use `@BasePathAwareController` but there seems to an issue with mapping the `basePath` into underlying method request mappings.
Also, using `@RequestMapping("${spring.data.rest.base-path:/api}")` has issues outside the servlet context.

ilayaperumalg:
Rebased.

spring-cloud-skipper-server/src/main/resources/application.yml
```diff
@@ -7,6 +7,9 @@ server:
 spring:
   main:
     banner-mode: "off"
+  data:
+    rest:
+      base-path: /api
```
markpollack: shoudl we make this `/api/repositories` ?

ghillert: This is just the base-path, `repositories` will indeed be available at `/api/repositories`

ghillert:
LGTM. Merged with commit aad93881899cd6348ebc04791dc5b5ec6eeef567

PR #157, opened 2017-10-09
Move exception from server to spring-cloud-skipper

### LINKED ISSUE #146: Move SkipperException to common module
This will enable sharing between client/server.  also move ReleaseNotfoundException to the domain module.  Can remove the `ExceptionHandler` in SkipperController for ReleaseNotFound, this should make for better error handling in the short term before the VndError approach is adopted.

jvalkeal:
- As these exceptions are not only server related, move
  to common shared package so that we can use same
  classes in a client as well. This would be to
  i.e. propagate ReleaseNotFoundException from server
  to client via rest.
- Made some static methods in ReleaseNotFoundException
  private as if those are only accessed in tests then
  its better to hardcode asserts in tests.
- Bit of polish and javadocs.
- Fixes #146


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as e87b2d0e1ecf4f8ad5bd711261521de5e55d81b2

PR #163, opened 2017-10-10
Fix assignment of groupname when deploying

### LINKED ISSUE #153: Fix assignment of groupname when deploying
The current implementation is adding the version suffix into the group name. Instead the group name need to be the name of the application without the version suffix so that for spring cloud stream applications, consumer groups work as expected during an upgrade/rollback.

jvalkeal:
- Simply remove version identifier from group
  into a name of an app.
- There's really no tests to add and even
  manual testing from scdf is impossible
  at this moment. However I did check
  from boot env endpoint that all settings
  related to consumer partitions should remain same.
- Fixes #153

### DIFF
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeploymentRequestFactory.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeploymentRequestFactory.java
@@ -28,17 +28,33 @@ import org.springframework.core.io.Resource;
 import org.springframework.util.Assert;
 
 /**
- * Utilities for managing AppDeploymentRequests
+ * Factory managing {@link AppDeploymentRequest}s.
+ *
  * @author Mark Pollack
+ * @author Janne Valkealahti
  */
 public class AppDeploymentRequestFactory {
 
 	private final DelegatingResourceLoader delegatingResourceLoader;
 
+	/**
+	 * Instantiates a new {@code AppDeploymentRequestFactory}.
+	 *
+	 * @param delegatingResourceLoader the delegating resource loader
+	 */
 	public AppDeploymentRequestFactory(DelegatingResourceLoader delegatingResourceLoader) {
+		Assert.notNull(delegatingResourceLoader, "'delegatingResourceLoader' must be set");
 		this.delegatingResourceLoader = delegatingResourceLoader;
 	}
 
+	/**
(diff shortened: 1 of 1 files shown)


spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeploymentRequestFactory.java
```diff
@@ -60,8 +79,7 @@ public AppDeploymentRequest createAppDeploymentRequest(SpringBootAppKind springB
 			deploymentProperties.put(AppDeployer.COUNT_PROPERTY_KEY, String.valueOf(metadata.get("count")));
 		}
 
-		// TODO fix setting of group property with new version, will mess up consumer groups.
-		deploymentProperties.put(AppDeployer.GROUP_PROPERTY_KEY, releaseName + "-v" + version);
+		deploymentProperties.put(AppDeployer.GROUP_PROPERTY_KEY, releaseName);
```
ilayaperumalg: What happens in the case when the deployment properties has already set the `group` in it?

ilayaperumalg: At least in SCDF, we always set the `group` to the stream name. Do we want to override that here?

ilayaperumalg: when deploying via SCDF I believe we always want to use the group name to be the `streamName` as SCDF sets it every time the stream is deployed. Not sure if we want to override it with the `releaseName` when deploying via Skipper because, it potentially changes the group for each `release` the SCDF stream gets deployed/updated/rolledback.

Maybe we could only set the `group` if the deployment properties doesn’t have it set already?

markpollack: Yes, I think that is correct.  If the deployment group property is already set, as it will be when the package it uploaded to skipper from SCDF, we just use that.  Forgot to consider this case.  I'll make the change and merge.

markpollack:
merged as 5966a5f82988043ee16a397ef8a2cd71c2613e07 and 9bd6e9f6a920ec61ab0312239dde4050c7bab1a4  forgot to squash.

Issue #135, opened 2017-10-05
REST - Add missing endpoints to Api Root URL


ghillert:
E.g. currently the `/about` REST endpoint is not listed under `http://localhost:7577/`:

```json
{
  "_links" : {
    "appDeployerDatas" : {
      "href" : "http://localhost:7577/appDeployerDatas{?page,size,sort}",
      "templated" : true
    },
    "repositories" : {
      "href" : "http://localhost:7577/repositories{?page,size,sort}",
      "templated" : true
    },
    "deployers" : {
      "href" : "http://localhost:7577/deployers{?page,size,sort}",
      "templated" : true
    },
    "releases" : {
      "href" : "http://localhost:7577/releases{?page,size,sort}",
      "templated" : true
    },
    "packageMetadata" : {
      "href" : "http://localhost:7577/packageMetadata{?page,size,sort,projection}",
      "templated" : true
    },
    "profile" : {
      "href" : "http://localhost:7577/profile"
    }
  }
}
```

PR #172, opened 2017-10-11
gh-135 REST - Add missing endpoints to Api Root URL

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-skipper/issues/135

### DIFF
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
@@ -31,6 +31,7 @@ import org.springframework.cloud.skipper.domain.UploadRequest;
 import org.springframework.cloud.skipper.server.service.PackageService;
 import org.springframework.cloud.skipper.server.service.ReleaseService;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
@@ -103,9 +104,8 @@ public class SkipperController {
 	}
 
 	@RequestMapping(path = "/manifest/{name}", method = RequestMethod.GET)
-	@ResponseStatus(HttpStatus.OK)
-	public String manifest(@PathVariable("name") String name) {
-		return this.releaseService.manifest(name);
+	public ResponseEntity<String> manifest(@PathVariable("name") String name) {
+		return new ResponseEntity<String>(this.releaseService.manifest(name), HttpStatus.OK);
 	}
 
 	@RequestMapping(path = "/manifest/{name}/{version}", method = RequestMethod.GET)
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
@@ -109,9 +109,8 @@ public class SkipperController {
 	}
 
 	@RequestMapping(path = "/manifest/{name}/{version}", method = RequestMethod.GET)
-	@ResponseStatus(HttpStatus.OK)
-	public String manifest(@PathVariable("name") String name, @PathVariable("version") int version) {
-		return this.releaseService.manifest(name, version);
+	public ResponseEntity<String> manifest(@PathVariable("name") String name, @PathVariable("version") int version) {
+		return new ResponseEntity<String>(this.releaseService.manifest(name, version), HttpStatus.OK);
 	}
 
 	@RequestMapping(path = "/upgrade", method = RequestMethod.POST)
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/index/SkipperControllerResourceProcessor.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/index/SkipperControllerResourceProcessor.java
@@ -34,13 +34,17 @@ public class SkipperControllerResourceProcessor implements ResourceProcessor<Rep
 		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).getAboutInfo()).withRel("about"));
 		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).upload(null)).withRel("upload"));
 		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).install(null)).withRel("install"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).install(null, null)).withRel("install-with-package-id"));
 		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).status(null)).withRel("status"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).status(null, 123)).withRel("status-by-name-and-version"));
 		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).manifest(null)).withRel("manifest"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).manifest(null, 123)).withRel("manifest-by-name-and-version"));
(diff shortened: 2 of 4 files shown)


spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
```diff
@@ -103,9 +104,8 @@ public Info status(@PathVariable("name") String name, @PathVariable("version") i
 	}
 
 	@RequestMapping(path = "/manifest/{name}", method = RequestMethod.GET)
-	@ResponseStatus(HttpStatus.OK)
-	public String manifest(@PathVariable("name") String name) {
-		return this.releaseService.manifest(name);
+	public ResponseEntity<String> manifest(@PathVariable("name") String name) {
```
ilayaperumalg: What is the reasoning behind changing the type here? This change may require changes to the client side as well.

ghillert: This is interesting - It looks like this does not work on pure String:

`methodOn(SkipperController.class).manifest(null)).withRel("manifest")`, since String is final. 

The recommendation seems to be to return a `ResponseEntity` instead. It should not change anything for the client, no?

ilayaperumalg: I can check if this works on the client.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/index/SkipperControllerResourceProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+@Component
+public class SkipperControllerResourceProcessor implements ResourceProcessor<RepositoryLinksResource> {
+
+	@Override
+	public RepositoryLinksResource process(RepositoryLinksResource resource) {
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).getAboutInfo()).withRel("about"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).upload(null)).withRel("upload"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).install(null)).withRel("install"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).status(null)).withRel("status"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).manifest(null)).withRel("manifest"));
```
ilayaperumalg: Shouldn't we add other REST endpoints like `/manifest/{name}/{version}` and `/list/{name}` etc., ?

ilayaperumalg: @ghillert what do you think about this?

ghillert:
@ilayaperumalg Rebased + Added missing links

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/index/SkipperControllerResourceProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+	@Override
+	public RepositoryLinksResource process(RepositoryLinksResource resource) {
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).getAboutInfo()).withRel("about"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).upload(null)).withRel("upload"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).install(null)).withRel("install"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).install(null, null)).withRel("install-with-package-id"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).status(null)).withRel("status"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).status(null, 123)).withRel("status-by-name-and-version"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).manifest(null)).withRel("manifest"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).manifest(null, 123)).withRel("manifest-by-name-and-version"));
```
ilayaperumalg: Are these names such as `manifest-by-name-and-version`, `install-with-package-id` and `status-by-name-and-version` used somewhere else to access these REST endpoints or are they just the names given to these links?

spring-cloud-skipper-server/src/test/java/org/springframework/cloud/skipper/server/controller/docs/ApiDocumentation.java
```diff
@@ -68,12 +67,28 @@ public void errors() throws Exception {
...
 			linkWithRel("appDeployerDatas").description("Exposes App Deployer Data"),
 			linkWithRel("repositories").description("Exposes package repositories"),
 			linkWithRel("deployers").description("Exposes deployer"),
 			linkWithRel("releases").description("Exposes release information"),
 			linkWithRel("packageMetadata").description("Provides details for Package Metadata"),
-			linkWithRel("profile").description("Entrypoint to provide ALPS metadata defining simple descriptions of application-level semantics.")
+			linkWithRel("profile").description("Entrypoint to provide ALPS metadata defining simple descriptions of application-level semantics"),
+
+			linkWithRel("status").description("Status for a last known release version"),
+			linkWithRel("status-by-name-and-version").description("Status for a last known release version by release name"),
```
ilayaperumalg: This needs to be for the release of the given name and version. Will fix while merging.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/index/SkipperControllerResourceProcessor.java
ilayaperumalg: To be consistent with the URIs, I think it would be better specify the URI variables as part of the relations values. For instance, we can use `/manifest/name/version` instead of `manifest-by-name-and-version`. I see we use the same approach at SCDF as well. 
I think we can use the same here as well.  I can update those changes while merging.

spring-cloud-skipper-server/src/test/java/org/springframework/cloud/skipper/server/controller/docs/ApiDocumentation.java
```diff
@@ -68,12 +67,28 @@ public void errors() throws Exception {
...
+			linkWithRel("install-with-package-id").description("Install a package by also providing the package id"),
 			linkWithRel("appDeployerDatas").description("Exposes App Deployer Data"),
 			linkWithRel("repositories").description("Exposes package repositories"),
 			linkWithRel("deployers").description("Exposes deployer"),
 			linkWithRel("releases").description("Exposes release information"),
 			linkWithRel("packageMetadata").description("Provides details for Package Metadata"),
-			linkWithRel("profile").description("Entrypoint to provide ALPS metadata defining simple descriptions of application-level semantics.")
+			linkWithRel("profile").description("Entrypoint to provide ALPS metadata defining simple descriptions of application-level semantics"),
+
+			linkWithRel("status").description("Status for a last known release version"),
```
ilayaperumalg: by the given release name.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/index/SkipperControllerResourceProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).upgrade(null)).withRel("upgrade"));
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(SkipperController.class).rollback(null, -1)).withRel("rollback"));
```
ilayaperumalg: This may not be `-1` as the rollback version can also take typical version int.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squshed,  rebased and merged as cb6f31ced48a706d824839f3af15619719549691 along with the changes that address the latest review comments 1354870f744f5ee1265f54e769d7665c4681a46e

PR #181, opened 2017-10-12
Remove component scan

### LINKED ISSUE #134: Remove component scan
In favour of having proper control of configs and beans, remove scanning and use combination of auto-configs and imported configs.

All this makes life easier when creating proper tests and not trying to test whole app as a black box.

jvalkeal:
- Similarly as in scdf server, introduce EnableSkipperServer,
  its Marker auto-config bean with EnableSkipperServerConfiguration
  and SkipperServerAutoConfiguration which imports SkipperServerConfiguration
  which have manually defined beans which used to be scanned.
- Fix tests which relied on full auto-config. These tests now
  add only needed auto-config classes and further tests
  can have better control what tests actually do.
- SkipperServerApplication was moved under `app` package for
  scan not to happen with @SpringBootApplication. This now uses
  @EnableSkipperServer.
- NOTE: SkipperServerApplication should be in its own module which
  only produces a fat-jar. I didn't want to do this now as it needs
  a rename of `spring-cloud-skipper-server` to `spring-cloud-skipper-server-core`
  causing impossible review as every path changes. When this rename is done
  in future, then only SkipperServerApplication stays in `spring-cloud-skipper-server`.
- Fixes #134

### DIFF
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/app/SkipperServerApplication.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/app/SkipperServerApplication.java
@@ -13,24 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.springframework.cloud.skipper.server;
+package org.springframework.cloud.skipper.server.app;
 
 import org.springframework.boot.SpringApplication;
-import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.cloud.deployer.spi.cloudfoundry.CloudFoundryDeployerAutoConfiguration;
 import org.springframework.cloud.deployer.spi.kubernetes.KubernetesAutoConfiguration;
 import org.springframework.cloud.deployer.spi.local.LocalDeployerAutoConfiguration;
-import org.springframework.data.map.repository.config.EnableMapRepositories;
+import org.springframework.cloud.skipper.server.EnableSkipperServer;
 
-@SpringBootApplication
-@EnableAutoConfiguration(exclude = { CloudFoundryDeployerAutoConfiguration.class,
-		LocalDeployerAutoConfiguration.class, KubernetesAutoConfiguration.class })
-@EnableMapRepositories
+@SpringBootApplication(exclude = { CloudFoundryDeployerAutoConfiguration.class, LocalDeployerAutoConfiguration.class,
+		KubernetesAutoConfiguration.class })
+@EnableSkipperServer
 public class SkipperServerApplication {
 
 	public static void main(String[] args) {
 		SpringApplication.run(SkipperServerApplication.class, args);
 	}
-
 }
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerAutoConfiguration.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerAutoConfiguration.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.skipper.server.config;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+/**
+ * Auto-configuration for skipper server.
+ *
+ * @author Janne Valkealahti
+ */
+@Configuration
+@ConditionalOnBean(EnableSkipperServerConfiguration.Marker.class)
+@Import(SkipperServerConfiguration.class)
+@ConditionalOnProperty(prefix = "spring.cloud.skipper.server", name = "enabled", havingValue = "true", matchIfMissing = true)
+public class SkipperServerAutoConfiguration {
+}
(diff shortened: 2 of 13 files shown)


spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
```diff
@@ -46,20 +49,40 @@
...
  * @author Ilayaperumal Gopinathan
+ * @author Janne Valkealahti
  */
 @Configuration
 @EnableConfigurationProperties({ SkipperServerProperties.class, CloudFoundryPlatformProperties.class,
 		LocalPlatformProperties.class, KubernetesPlatformProperties.class,
 		MavenConfigurationProperties.class })
 @EntityScan({"org.springframework.cloud.skipper.domain", "org.springframework.cloud.skipper.server.domain"})
+@EnableMapRepositories(basePackages = "org.springframework.cloud.skipper.server.repository")
+@EnableJpaRepositories(basePackages = "org.springframework.cloud.skipper.server.repository")
```
ilayaperumalg: We didn't have `EnableJpaRepositories` before. Is this now added for fixing any specific issue?

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerAutoConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Auto-configuration for skipper server.
+ *
+ * @author Janne Valkealahti
+ */
+@Configuration
+@ConditionalOnBean(EnableSkipperServerConfiguration.Marker.class)
+@Import(SkipperServerConfiguration.class)
+@ConditionalOnProperty(prefix = "spring.cloud.skipper.server", name = "enabled", havingValue = "true", matchIfMissing = true)
```
ilayaperumalg: We have this property `spring.cloud.skpper.server.enabled` in SCDF to enable Skipper but I'd like to understand this property's usage inside Skipper itself given we already have `EnableSkipperServer` as an annotation.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/app/SkipperServerApplication.java
```diff
@@ -13,24 +13,21 @@
...
 import org.springframework.cloud.deployer.spi.kubernetes.KubernetesAutoConfiguration;
 import org.springframework.cloud.deployer.spi.local.LocalDeployerAutoConfiguration;
-import org.springframework.data.map.repository.config.EnableMapRepositories;
+import org.springframework.cloud.skipper.server.EnableSkipperServer;
 
-@SpringBootApplication
-@EnableAutoConfiguration(exclude = { CloudFoundryDeployerAutoConfiguration.class,
-		LocalDeployerAutoConfiguration.class, KubernetesAutoConfiguration.class })
-@EnableMapRepositories
+@SpringBootApplication(exclude = { CloudFoundryDeployerAutoConfiguration.class, LocalDeployerAutoConfiguration.class,
```
ilayaperumalg: It might be better to exclude these autoconfiguration classes inside the config (maybe the application.yml) rather than the main application here. That way, a custom main application with `EnableSkipperServer` would still need not exclude these autoconfiguration classes explicitly.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerAutoConfiguration.java
jvalkeal: just a way to disable auto-config if needed.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
jvalkeal: Didn't try any other way as you need to tell where repos as scanned. Previously full boot auto-config did the scanned from app itself.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/app/SkipperServerApplication.java
jvalkeal: Nope, we can't. You can't exclude anything within a config itself as that will just create a mess. Exclude needs to happen on a top level component(SkipperServerApplication) or with a property(we disabled SessionAutoConfiguration) in scdf that way. I'm more interested to know why those auto-configs had to be excluded.

ilayaperumalg: we want to have control over creating the deployer configuration and that's why we disable these deployer auto configurations explicitly.

jvalkeal: we should improve their auto-config so that we wouldn't need to disable.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
ilayaperumalg: I suspect we may not need this but we can verify this.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/app/SkipperServerApplication.java
ilayaperumalg: yeah, I agree that the deployer auto configurations are flexible enough to configure enable/disable based on some conditions. In the meantime, I think we are OK having this (with just a note that every main application needs to disable these auto configurations explicitly).

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
ilayaperumalg: will verify this while merging.

ilayaperumalg:
LGTM, merging.

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
ilayaperumalg: I could confirm the `EnableJpaRepositories` is indeed needed :-)

ilayaperumalg:
Rebased and merged as 5142c715a4890e1876674555c0f6c75c759b8e06

PR #184, opened 2017-10-13
Check if package exists before update

### LINKED ISSUE #179: update needs to guard against not finding package metadata
In the update method
`packageMetadataRepository		.findByNameAndVersion`

can return null.  Add a null check (as in install) or add another custom repository method that will throw an exception.  findByNameAndVersionRequired...

ilayaperumalg:
- Throw SkipperException if the package with the name/version doesn't exist
 - Refactor the existing check in install and re-use it for the upgrade as well
 - Add tests

Resolves #179


jvalkeal:
lgtm

jvalkeal:
squashed and merged per https://github.com/spring-cloud/spring-cloud-skipper/commit/aef45d3e3be2699e4a02c7c1a212c539521e589c

PR #185, opened 2017-10-13
Fix search query in packageMetadata repo

### LINKED ISSUE #177: search command is not doing basic 'like' matching for package names
`search --name tim` does not return time.

ilayaperumalg:
- Use `findByNameContainingIgnoreCase` instead of `findByNameLike`

Resolves #177


jvalkeal:
lgtm

PR #194, opened 2017-10-16
Reuse deleted release history

### LINKED ISSUE #193: Support installing over a previously deleted release name
As a user, I'd like to re-create the release with the same name after it is marked `deleted` from the previous history.

ilayaperumalg:
- If the release is deleted, allow creation of subsequent releases with the same release name. This way, the releases with the previously deleted one can be re-created and the history is preserved as well.
 - Add custom release repository method to get the latest release by the given name and the status of the release is `deleted`. This can further be used by the subsequent release creation methods to increment the release versions
 - Implement the repository method
 - Update ReleaseService to check for previously deleted releases by the given name and use the last revision version to calculate the currently installed version
 - Add tests for release repository and Release service methods

Resolves #193

### DIFF
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/SimpleRedBlackUpgradeStrategy.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/SimpleRedBlackUpgradeStrategy.java
@@ -85,8 +85,7 @@ public class SimpleRedBlackUpgradeStrategy implements UpgradeStrategy {
 
 		// Carry over the applicationDeployment information for apps that were not updated.
 		AppDeployerData existingAppDeployerData = this.appDeployerDataRepository.findByReleaseNameAndReleaseVersion(
-				existingRelease.getName(),
-				existingRelease.getVersion());
+				existingRelease.getName(), existingRelease.getVersion());
 		Map<String, String> existingAppNamesAndDeploymentIds = existingAppDeployerData.getDeploymentDataAsMap();
 
 		for (Map.Entry<String, String> existingEntry : existingAppNamesAndDeploymentIds.entrySet()) {
--- a/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/repository/ReleaseRepositoryImpl.java
+++ b/spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/repository/ReleaseRepositoryImpl.java
@@ -86,4 +86,11 @@ public class ReleaseRepositoryImpl implements ReleaseRepositoryCustom {
 		}
 		return releases;
 	}
+
+	@Override
+	public Release findLatestReleaseIfDeleted(String releaseName) {
+		Release latestRelease = this.releaseRepository.findTopByNameOrderByVersionDesc(releaseName);
+		return (latestRelease != null &&
+				latestRelease.getInfo().getStatus().getStatusCode().equals(StatusCode.DELETED)) ? latestRelease : null;
+	}
 }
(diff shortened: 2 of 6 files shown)


spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/SimpleRedBlackUpgradeStrategy.java
```diff
@@ -85,8 +85,7 @@ public Release upgrade(Release existingRelease, Release replacingRelease,
 
 		// Carry over the applicationDeployment information for apps that were not updated.
 		AppDeployerData existingAppDeployerData = this.appDeployerDataRepository.findByReleaseNameAndReleaseVersion(
-				existingRelease.getName(),
-				existingRelease.getVersion());
+				existingRelease.getName(), existingRelease.getVersion());
```
markpollack: I think we still have some intellij formatting diffs...

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/repository/ReleaseRepositoryImpl.java
```diff
@@ -86,4 +86,11 @@ public Release findByNameAndVersion(String releaseName, int version) {
 		}
 		return releases;
 	}
+
+	@Override
+	public Release findLatestReleaseIfDeleted(String releaseName) {
+		Release latestRelease = this.releaseRepository.findTopByNameOrderByVersionDesc(releaseName);
+		return (latestRelease != null &&
+				latestRelease.getInfo().getStatus().getStatusCode().equals(StatusCode.DELETED)) ? latestRelease : null;
```
markpollack: I wonder if we need to handle 'FAILED' as well.  We need to review the states in StatusCode....

spring-cloud-skipper-server/src/main/java/org/springframework/cloud/skipper/server/service/ReleaseService.java
```diff
@@ -167,7 +171,10 @@ private void validateInstallRequest(InstallRequest installRequest) {
 
 	protected Release install(PackageMetadata packageMetadata, InstallProperties installProperties) {
 		Assert.notNull(packageMetadata, "Can't download package, PackageMetadata is a null value.");
-		Release release = createInitialRelease(installProperties, this.packageService.downloadPackage(packageMetadata));
+		Release existingDeletedRelease = this.releaseRepository.findLatestReleaseIfDeleted(installProperties.getReleaseName());
```
markpollack: I don't think we want this behavior for install.  The first use-case, is to be able to rollback a release that is deleted.  That seems analogous to the 'trash can' concept in an OS, we can restore something we put in the trash.

In helm, there is a specific option wrt to supporting this on install, the option `--replace`

From the helm description of that install option, `re-use the given name, even if that name is already used. This is unsafe in production`

I do wonder the motivation around that last comment, not safe in production.  I believe doing a `rollback` followed by an `upgrade` would be the logical equivalent.

Created https://github.com/spring-cloud/spring-cloud-skipper/issues/207 to track

markpollack: I will add some info logging for now to indicate which code pathway is being taken.

spring-cloud-skipper-server/src/test/java/org/springframework/cloud/skipper/server/service/ReleaseServiceTests.java
```diff
@@ -126,12 +126,55 @@ public void testStatusReleaseExist() {
...
+			fail("Expected to fail when installin already deployed release.");
+		}
+		catch (SkipperException e) {
+			assertThat(e.getMessage()).isEqualTo("Release with the name [" + releaseName + "] already exists "
+					+ "and it is not deleted.");
+		}
+	}
+
+	@Test
+	public void testInstallDeletedRelease() {
```
markpollack: Maybe it was out of scope for this PR, driven by the SCDF deploy an already deleted stream use-case, but I will add a test (and any changed) to support rolling back a deleted release.

PR #204, opened 2017-10-17
Calculate app count based on runtime status

### LINKED ISSUE #152: Take into account the current number of deployed applications when upgrading
The count value in the package metadata is only used for the initial install.  When upgrading or rolling back, the target count should be equal to the currently executing count.

This will not work on the local deployer, but it should work on PCF and K8s.  Need to be checked on those platforms.

### LINKED ISSUE #199: Remove metadata.count field in template
The value of count should only come from the deployer properties.  This was put in there by mistake in trying to mimic a k8s style yaml resource document (in which case it should have gone directly under `spec` ala `replicaCount: 3`.

There is code that is now copying the deployment properties to the metadata:count field, should be removed.

ilayaperumalg:
- When upgrade/rollback, the application count needs to be discovered based on the runtime application status
  - Calculate the number of applications based on the app status size and set it as the the `deploymentProperties` for the replacing release's manifest
  - Use constants for the keys

Fix package metadata to remove metadata.count attribute from the test packages

Resolves #152
Resolves #199


jvalkeal:
At least one error in tests(might be caused by previous docs PR from gunnar).
```
[ERROR] Errors: 
[ERROR] org.springframework.cloud.skipper.server.controller.docs.InstallDocumentation.installPackageWithId(org.springframework.cloud.skipper.server.controller.docs.InstallDocumentation)
[ERROR]   Run 1: InstallDocumentation.installPackageWithId:131 » NestedServlet Request processi...
[ERROR]   Run 2: InstallDocumentation>AbstractControllerTests.cleanupReleases:70 » NestedServlet
[INFO] 
[INFO] 
[ERROR] Tests run: 59, Failures: 0, Errors: 1, Skipped: 0
```

ilayaperumalg:
@jvalkeal That test error was because of the removal `spring.cloud.deployer.group` as deployment property if not set already. Fixed that and updated the PR.

jvalkeal:
I think this looks ok.

jvalkeal:
squashed/merged per https://github.com/spring-cloud/spring-cloud-skipper/commit/e2dd4536811eeae893f1ef736fc879323cd7db8b

PR #206, opened 2017-10-17
Fix client rest error handling

### LINKED ISSUE #173: SkipperStreamDeployer state mapping when ReleseNotFoundException is thrown
When creating a new stream and not calling `--deploy` but using `--useSkipper` a subsequent call to `stream status` will return a `ReleaseNotFoundException` from skipper. This can be fixed by catching this exception and setting the deployment state of the stream to unknown.  

Also, now that ReleaseNotFoundException is in the core shared project between skipper clients and server, we can remove 

```
	@ResponseStatus(value = HttpStatus.NOT_FOUND, reason = "Release doesn't exist")
	@ExceptionHandler(ReleaseNotFoundException.class)
	public void releaseNotExist() {
		// handle ReleaseNotFoundException by returning 404
	}```

in SkipperController.

### LINKED ISSUE #192: Wrong RestTemplate configuration
`SkipperClientConfiguration` creates `restTemplate` as a bean which is then used as a dependency with `skipperClient`. 

Trouble is that in `TargetHolder`, a new `DefaultSkipperClient` is created manually with a new `RestTemplate` effectively making `SkipperClientConfiguration` useless.

### PR DESCRIPTION
**Pull request by jvalkeal**
- Add custom SkipperErrorAttributes for getting
  skipper related errors for json object.
- Add custom SkipperClientResponseErrorHandler into
  RestTemplate to throw ReleaseNotFoundException
  with releaseName and releaseVersion if those exists
  in error object.
- Fix TargetHolder to use RestTemplate created in
  SkipperClientConfiguration.
- Fixes #192
- Fixes #173

### DIFF
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/SkipperClientResponseErrorHandler.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/SkipperClientResponseErrorHandler.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.skipper.client;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.springframework.cloud.skipper.ReleaseNotFoundException;
+import org.springframework.cloud.skipper.SkipperException;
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/SkipperClientResponseErrorHandler.java [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+		Map<String, String> map = new HashMap<>();
+		try {
+			String body = new String(getResponseBody(response));
+			@SuppressWarnings("unchecked")
+			Map<String, String> parsed = objectMapper.readValue(body, Map.class);
+			map.putAll(parsed);
+			exceptionClazz = map.get("exception");
+		}
+		catch (Exception e) {
+			// don't want to error here
```
markpollack: shouldn't we log some message that we couldn't read the expected response from the server?

markpollack:
also tested with 
```
skipper:>install --release-name mylog --package-name log --package-version 2.0.0
Released mylog
skipper:>status --release-name asdf
Release with name 'asdf' not found
```
which corresponds to 
```
		Info info;
		try {
			if (releaseVersion == null) {
				info = this.skipperClient.status(releaseName);
			}
			else {
				info = this.skipperClient.status(releaseName, releaseVersion);
			}
		}
		catch (ReleaseNotFoundException e) {
			return "Release with name '" + e.getReleaseName() + "' not found";
		}
```
in the status command on the client side.

markpollack:
Fixed small typo and merged as 6656e6be374facc7ac623d79f7853c7faffa05bd

PR #210, opened 2017-10-18
Throws SkipperException vs IncorrectResultSizeDataAccessException

### LINKED ISSUE #195: Throw exception if same name package is present in two different repositories
In a later release, will handle a fallback mechanism based on the order property of the repo.

cppwfs:
Also did some cleanup in the tests

resolves #195


markpollack:
Created https://github.com/spring-cloud/spring-cloud-skipper/issues/212 to track resolving the issue when there is a same name/version package in multiple repos.

PR #214, opened 2017-10-19
gh-200 Change AbstractEntity's id from string to long

ghillert:
* Add `repositoryId` property to `PackageMetadata`
* Have service persist the `repositoryId` property instead of `origin` property

resolves https://github.com/spring-cloud/spring-cloud-skipper/issues/200


spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/PackageMetadata.java
```diff
@@ -35,12 +36,16 @@
 	private String apiVersion;
 
 	/**
-	 * The repository ID this Package Index file belongs to. todo: Make this as Required once
-	 * we have repository backed package for upload etc.,
+	 * Indicates the origin of the repository (free form text).
 	 */
 	private String origin;
```
ilayaperumalg: origin is the repository Id and hence we may not need `repositoryId`

ilayaperumalg: I might be missing something here. What is the intention behind supporting both the `repositoryId` as well as the `origin` while we only had the `origin` pointing to the repository id before

jvalkeal: looking y'day discussion on slack, per mark's comments
```
If we add a new property to packageMetadata such as 'repositoryId', then that woudl be set here instead.
and in the other direction `Repository packageRepository = repositoryRepository.findOne(packageMetadata.getOrigin());`
would change to `Repository packageRepository = repositoryRepository.findOne(packageMetadata.getRepositoryId());`
```
looks like he meant replace, not just add new one?

ilayaperumalg:
LGTM, merging

ilayaperumalg:
Rebased, resolved the conflicts and merged as eccaca8aa12e032ee84f32fbba57abe323f449f2 along with afd10907ae0a57576dbb49ba32e41e397c4576b7 that removes the unused property PackageMetadata.origin and its references in tests and documentation.

PR #216, opened 2017-10-19
Search package metadata across repositories

### LINKED ISSUE #212: Resolve ambiguities in releaseService.getPackageMetadata when multiple packages are in different repos
The `order` property in the repository should be used to return a single packagemetadata object based on which repository has higher priority, e.g. order.

ilayaperumalg:
- When finding package metadata by name and version, following algorithm is used:
  - If there is a single package metadata found, then return
  - If there are multiple package metadata, then return the package metadata that belongs to the higher order repository
  - If there are no repository order information found, then return the package metadata that has the higher api version
  - Add custom repository method in PackageMetadataRepository and RepositoryRepository
  - Add tests

Resolves #212


jvalkeal:
lgtm

Issue #196, opened 2017-10-16
Add @Version field to AbstractEntity


markpollack:
Add

`private @Version Long version;`

for optimistic locking.

PR #221, opened 2017-10-19
gh-196 Add @Version field to AbstractEntity

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-skipper/issues/196

### DIFF
--- a/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/AbstractEntity.java
+++ b/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/AbstractEntity.java
@@ -19,6 +19,7 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.MappedSuperclass;
+import javax.persistence.Version;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/AbstractEntity.java
```diff
@@ -38,6 +39,10 @@
 	@JsonIgnore
 	private final Long id;
 
+	@Version
+	@JsonIgnore
+	private Long objectVersion;
```
cppwfs: nitpicky... I know... Why objectVersion instead of just version?

cppwfs: Nevermind.  Just read comment on it in Slack.

ghillert: `PackageMetadata` has a `version` (String) property.

ghillert:
This was merged with commit 9a1d0badae3b13689380f077caaa23d451b56e18

PR #231, opened 2017-10-20
Use latest package when version not specified

### LINKED ISSUE #219: On upgrade, pick latest package version, if not specified

ilayaperumalg:
- This is now applicable to both install and upgrade
 - Add test

Resolves #219


jvalkeal:
lgtm

PR #238, opened 2017-10-20
Upgrade properties parsing at client side

### LINKED ISSUE #235: Improve parsing of --properties at the client side when upgrading

ilayaperumalg:
- When parsing and loading YAML string for the upgrade properties YAML String, make sure both the `applicationProperties` and `deploymentProperties` are retrieved appropriately.
   - Since Props2Yml consider each dotted value as a separate key, hack the spec properties so that they don't get considered as the spec property keys

Resolves #235


markpollack:
Made some of the static strings private and squahsed/merged as f1eefafcb57bf0d0dfd1f18c34a99b19e5121ce1

Issue #149, opened 2017-10-08
Upload skipper docker image to dockerhub in CI plan


markpollack:

ilayaperumalg:
I believe this can be closed as we publish the docker images into dockerhub via CI plan now.

sabbyanandan:
Yes - it is complete.

Issue #148, opened 2017-10-08
Add Docker build mvn build


markpollack:
For kubernetes, there need to be docker build to deploy skipper.

sabbyanandan:
This is addressed in: https://github.com/spring-cloud/spring-cloud-skipper/commit/e5c52a2ee715a2eb75baa83f4e11d228ecb98d39.

PR #241, opened 2017-10-22
Add support for build and push docker image

eddumelendez:
Resolves spring-cloud/spring-cloud-skipper#148
Resolves spring-cloud/spring-cloud-skipper#149

### DIFF
--- a/spring-cloud-skipper-server/Dockerfile
+++ b/spring-cloud-skipper-server/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8u141
+FROM java:8-alpine
 
 ARG JAR_FILE
 
--- a/spring-cloud-skipper-server/Dockerfile
+++ b/spring-cloud-skipper-server/Dockerfile
@@ -1,0 +1,7 @@
+FROM openjdk:8u141
+
+ARG JAR_FILE
+
+ADD target/${JAR_FILE} spring-cloud-skipper-server.jar
+
+ENTRYPOINT exec java -Djava.security.egd=file:/dev/./urandom -jar /spring-cloud-skipper-server.jar
--- a/spring-cloud-skipper-server/pom.xml
+++ b/spring-cloud-skipper-server/pom.xml
@@ -12,7 +12,7 @@
     </parent>
 
     <properties>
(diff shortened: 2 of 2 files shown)


markpollack:
Thanks so much!

spring-cloud-skipper-server/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openjdk:8u141
```
trisberg: We use `java:8-alpine` for Data Flow images

spring-cloud-skipper-server/pom.xml
```diff
@@ -11,6 +11,10 @@
         <version>1.0.0.BUILD-SNAPSHOT</version>
     </parent>
 
+    <properties>
+        <docker.image.prefix>spring-io</docker.image.prefix>
+    </properties>
```
trisberg: could we use `springcloud` here, that's what we use for Data Flow

spring-cloud-skipper-server/pom.xml
```diff
@@ -53,4 +57,43 @@
             </plugin>
         </plugins>
     </build>
+
+	<profiles>
+		<profile>
+			<id>docker</id>
+
```
trisberg: do we want this as a profile or just use `./mvnw package dockerfile:build` without adding the plugin to the package phase

eddumelendez:
@trisberg thanks for feedback, I have update the docker image and the prefix in the new commit. Regarding to the docker profile I will wait until the decision is taken.

spring-cloud-skipper-server/pom.xml
sabbyanandan: Probably just fine with `dockerfile:build` just to be consistent with other projects that generate docker images.

sabbyanandan:
Thanks for the update, @eddumelendez! See my comment re: profile ^^. Hopefully, when you're done with that change, we can go ahead and merge it. 😄

eddumelendez:
change is done 😃

trisberg:
LGTM

trisberg:
Rebased and merged as commit e5c52a2ee715a2eb75baa83f4e11d228ecb98d39

Issue #207, opened 2017-10-17
Investigate adding a `--replace` option on install


markpollack:
With https://github.com/spring-cloud/spring-cloud-skipper/pull/194/ we can do an install over a release that was deleted.  This helps support a common development workflow from the Data Flow use case perspective, but we may want to introduce an explicit `--replace` option as in helm in order to use this functionality vs. not allowing an install to re-use an existing release name.

PR #246, opened 2017-10-23
Make RedBlackUpgrdeStrategy async

### LINKED ISSUE #244: Make RedBlackUpgradeStrategy asynchronous
Add health check of new releases which takes time,  up to several minutes on Cloud Foundry.

markpollack:
* Make SimpleRedBlackUpgradeStrategy asychronous for status check of new apps and delete of old
* Make SimpleRedBlackUpgradeStrategy a managed Bean
* Add configuration properties for timeout and sleep time of health check
* Break out SimpleRedBlackUpgradeStrategy into two steps
* Add AppDeployerDataRepository.findByReleaseNameAndReleaseVersionRequired to avoid NPEs
* Check for fast fail on route-path deployment property for CF not having leading 'slash' /
* Catch exception inside for loop when deleting releases on test cleanup
* Update tests to deal with async behavior

Fixes #244

Unfortunately increases test time, maybe RESTDocs can use mocks.

Running Skipper with this config file
```
spring:
  cloud:
    skipper:
      server:
        synchonizeIndexOnContextRefresh: true      
        packageRepositories:
          -
            name: experimental
            url: "http://localhost:8080/repository/experimental"
            description: Sample Skipper Repository      
          -
            name: test
            url: "file:///home/mpollack/projects/spring-cloud-skipper/spring-cloud-skipper-server-core/src/test/resources/repositories/binaries/test/"
          -
            name: local
            url: "http://localhost:7577"
            local: true
            description: Default local database backed repository
        platform:
          cloudfoundry:
            accounts:
              cf-dev:
                connection:
                  url: https://api.run.pivotal.io
                  org: scdf-ci
                  space: space-mark
                  domain: cfapps.io
                  username: your@email.com
                  password: xxxx
                  skipSslValidation: false
                deployment:
                  deleteRoutes: false
```

Of note is the `deployment.deleteRoutes = false`.  That is important to have a seamless transition to the new app when hitting the http route during the transition.

I deployed as such 
```
install --release-name myhelloworld --package-name helloworld --package-version 1.0.0 --platform-name cf-dev --file /home/mpollack/helloworld-deployment.yml
```
Where the file `helloworld-deployment.yml` was
```
spec:
  deploymentProperties:
    spring.cloud.deployer.cloudfoundry.route: mlp-helloworld.cfapps.io
```
and then upgraded
```
upgrade --release-name myhelloworld --package-name helloworld --package-version 1.0.0 --file /home/mpollack/helloworld-upgrade.yml
```
where `helloworld-upgrade.yml` was
```
 spec:
  applicationProperties:
    helloworld.greeting: yo
  deploymentProperties:
    spring.cloud.deployer.cloudfoundry.route: mlp-helloworld.cfapps.io
```

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
@@ -17,7 +17,12 @@ package org.springframework.cloud.skipper.server.config;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.Executor;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.domain.EntityScan;
 import org.springframework.boot.autoconfigure.web.ErrorAttributes;
@@ -36,6 +41,11 @@ import org.springframework.cloud.skipper.server.deployer.AppDeployerReleaseManag
 import org.springframework.cloud.skipper.server.deployer.AppDeploymentRequestFactory;
 import org.springframework.cloud.skipper.server.deployer.ReleaseAnalyzer;
 import org.springframework.cloud.skipper.server.deployer.ReleaseManager;
+import org.springframework.cloud.skipper.server.deployer.strategies.DeleteStep;
+import org.springframework.cloud.skipper.server.deployer.strategies.HealthCheckAndDeleteStep;
+import org.springframework.cloud.skipper.server.deployer.strategies.HealthCheckProperties;
+import org.springframework.cloud.skipper.server.deployer.strategies.SimpleRedBlackUpgradeStrategy;
+import org.springframework.cloud.skipper.server.deployer.strategies.UpgradeStrategy;
 import org.springframework.cloud.skipper.server.index.PackageMetadataResourceProcessor;
 import org.springframework.cloud.skipper.server.index.PackageSummaryResourceProcessor;
 import org.springframework.cloud.skipper.server.index.SkipperControllerResourceProcessor;
@@ -55,6 +65,9 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
 import org.springframework.data.map.repository.config.EnableMapRepositories;
+import org.springframework.scheduling.annotation.AsyncConfigurer;
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
 
 /**
@@ -68,12 +81,15 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
 @Configuration
 @EnableConfigurationProperties({ SkipperServerProperties.class, CloudFoundryPlatformProperties.class,
 		LocalPlatformProperties.class, KubernetesPlatformProperties.class,
-		MavenConfigurationProperties.class })
+		MavenConfigurationProperties.class, HealthCheckProperties.class })
 @EntityScan({ "org.springframework.cloud.skipper.domain", "org.springframework.cloud.skipper.server.domain" })
 @EnableMapRepositories(basePackages = "org.springframework.cloud.skipper.server.repository")
 @EnableJpaRepositories(basePackages = "org.springframework.cloud.skipper.server.repository")
 @EnableTransactionManagement
-public class SkipperServerConfiguration {
+@EnableAsync
+public class SkipperServerConfiguration implements AsyncConfigurer {
+
+	private final Logger logger = LoggerFactory.getLogger(SkipperServerConfiguration.class);
 
 	@Bean
 	public ErrorAttributes errorAttributes() {
@@ -172,9 +188,51 @@ public class SkipperServerConfiguration {
 			AppDeployerDataRepository appDeployerDataRepository,
 			DeployerRepository deployerRepository,
 			ReleaseAnalyzer releaseAnalyzer,
-			AppDeploymentRequestFactory appDeploymentRequestFactory) {
+			AppDeploymentRequestFactory appDeploymentRequestFactory,
+			UpgradeStrategy updateStrategy) {
 		return new AppDeployerReleaseManager(releaseRepository, appDeployerDataRepository, deployerRepository,
-				releaseAnalyzer, appDeploymentRequestFactory);
+				releaseAnalyzer, appDeploymentRequestFactory, updateStrategy);
+	}
+
+	@Bean
+	public DeleteStep deleteStep(ReleaseRepository releaseRepository,
+			DeployerRepository deployerRepository,
+			AppDeployerDataRepository appDeployerDataRepository) {
+		return new DeleteStep(releaseRepository, deployerRepository, appDeployerDataRepository);
+	}
+
+	@Bean
+	public UpgradeStrategy updateStrategy(ReleaseRepository releaseRepository,
+			DeployerRepository deployerRepository,
+			AppDeployerDataRepository appDeployerDataRepository,
+			AppDeploymentRequestFactory appDeploymentRequestFactory,
+			HealthCheckAndDeleteStep healthCheckAndDeleteStep) {
+		return new SimpleRedBlackUpgradeStrategy(releaseRepository, deployerRepository,
+				appDeployerDataRepository, appDeploymentRequestFactory, healthCheckAndDeleteStep);
+	}
+
+	@Bean
+	public HealthCheckAndDeleteStep healthCheckAndDeleteStep(ReleaseRepository releaseRepository,
+			DeployerRepository deployerRepository,
+			AppDeployerDataRepository appDeployerDataRepository,
+			DeleteStep deleteStep,
+			HealthCheckProperties healthCheckProperties) {
+		return new HealthCheckAndDeleteStep(releaseRepository,
+				deployerRepository,
+				deleteStep,
+				healthCheckProperties);
+	}
+
+	@Bean(name = "skipperThreadPoolTaskExecutor")
+	@Override
+	public Executor getAsyncExecutor() {
+		return new ThreadPoolTaskExecutor();
+	}
+
+	@Override
+	public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
+		return (throwable, method, objects) -> logger.error("Exception thrown in @Async Method " + method.getName(),
+				throwable);
 	}
 
 	@Bean
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
@@ -26,12 +26,13 @@ import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.deployer.spi.app.AppDeployer;
 import org.springframework.cloud.deployer.spi.app.AppStatus;
+import org.springframework.cloud.deployer.spi.cloudfoundry.CloudFoundryDeploymentProperties;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 import org.springframework.cloud.skipper.SkipperException;
 import org.springframework.cloud.skipper.domain.Release;
 import org.springframework.cloud.skipper.domain.Status;
 import org.springframework.cloud.skipper.domain.StatusCode;
-import org.springframework.cloud.skipper.server.deployer.strategies.SimpleRedBlackUpgradeStrategy;
+import org.springframework.cloud.skipper.server.deployer.strategies.UpgradeStrategy;
 import org.springframework.cloud.skipper.server.domain.AppDeployerData;
 import org.springframework.cloud.skipper.server.domain.SpringBootAppKind;
 import org.springframework.cloud.skipper.server.domain.SpringBootAppKindReader;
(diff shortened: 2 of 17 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/DeleteStep.java
```diff
@@ -0,0 +1,88 @@
...
+				else {
+					logger.warn("For Release name {}, did not undeploy existing app {} as it status is not 'deployed'.",
+							release.getName(),
+							appNameAndDeploymentId.getKey());
+				}
+			}
+		}
+
+		Status deletedStatus = new Status();
+		deletedStatus.setStatusCode(StatusCode.DELETED);
```
ilayaperumalg: We might be marking the status to `DELETED` even for those who didn't have `deployed` state (the ones handled in else block above) ?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Reponsible for checking the health of the latest deployed release, and then deleting
+ * those applications from the previous release. Executed asynchronously to avoid the
+ * client waiting a long time, potentially minutes, when executing update.
+ * @author Mark Pollack
+ */
+public class HealthCheckAndDeleteStep {
+
+	private final Logger logger = LoggerFactory.getLogger(SimpleRedBlackUpgradeStrategy.class);
```
ilayaperumalg: Class name

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
```diff
@@ -0,0 +1,153 @@
...
+			DeployerRepository deployerRepository,
+			DeleteStep deleteStep,
+			HealthCheckProperties healthCheckProperties) {
+		this.releaseRepository = releaseRepository;
+		this.deployerRepository = deployerRepository;
+		this.deleteStep = deleteStep;
+		this.healthCheckProperties = healthCheckProperties;
+	}
+
+	@Async("skipperThreadPoolTaskExecutor")
```
ilayaperumalg: We could make this name a constant.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
```diff
@@ -0,0 +1,153 @@
...
+				replacingRelease.getInfo().setDescription("Upgrade complete");
+				this.releaseRepository.save(replacingRelease);
+				logger.info("Release {}-v{} has been DEPLOYED", replacingRelease.getName(),
+						replacingRelease.getVersion());
+			}
+			else {
+				logger.error("New release " + replacingRelease.getName() + " was not detected as healthy." +
+						"Keeping existing releases.");
+				Status status = new Status();
+				status.setStatusCode(StatusCode.DEPLOYED);
```
ilayaperumalg: Is the status supposed to be `DEPLOYED` even though the apps aren't healthy ?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
```diff
@@ -167,19 +217,16 @@ public Release upgrade(Release existingRelease, Release replacingRelease, String
 		return status(release);
```
ilayaperumalg: Now that the upgradeStrategy would asynchronously wait for apps to be deployed, but since the `upgrade` call returns and the `status(release)` may have issues getting the status for the release that is getting asynchronously checked?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/DeleteStep.java
markpollack: The way I interpret the status Code is that those apps for the release have been requested to be deleted.  The ones that carry over between releases done count.  We asked 'the platform' to delete the release.  That may have been one app or many, but the status of the release is 'requested deleted'. Does that make sense?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
markpollack: Yep, I think I need a macro for that....

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
markpollack: Yes, it has the potential to get out of sync

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
markpollack: Yes, we got rid of 'deploying' status code.  deployed is the state where we asked the platform to deploy our apps.  Similar to the deleted comment about.  I'm not super thrilled with the skipper status and platform status, but I don't see a better way ATM

markpollack: Yes, we got rid of 'deploying' status code.  deployed is the state where we asked the platform to deploy our apps.  Similar to the deleted comment about.  I'm not super thrilled with the skipper status and platform status, but I don't see a better way ATM

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
markpollack: I added kept the finder for AppDeploymentData to use the non-required version in status, in fact I ran into this race condition in testing.  It just returns the current release, with the idea the status method will be called again.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/DeleteStep.java
ilayaperumalg: yeah, it would make sense from the release point of view. But, once the release is marked `DELETED`, there could be subsequent releases to re-install them. In case where the platform didn't handle the deletion successfully, then the subsequent re-install operations would have apps from completely different package etc. based on the release's install properties.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
ilayaperumalg: Yeah, I see that. It makes sense :-)

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/DeleteStep.java
markpollack: If deleted fail (which I've actually never seen) there would still be running apps.  Any subsequent install operations for that release name would install whatever was requested.  In terms of it being completely different apps/packages - that problem exists in the sunny day scenario as well (delete succeedes).  The 'tis why I think there was the option in helm to by default not to be able to install and app that was deleted - I created an issue for us to investigate that more.

markpollack: https://github.com/spring-cloud/spring-cloud-skipper/issues/207

markpollack: The behavior was the same when this was synchronous...

ilayaperumalg: Right, this issue irrespective of sync vs async as it is all about what status we want to update the apps when they are not in already deployed state but deletion is requested.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
jvalkeal: At least with plain install calling `status` updates platform status to the `Release`.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
```diff
@@ -0,0 +1,153 @@
...
+			logger.error("Error upgrading new app", e);
+			// Update Status in DB
+			Status status = new Status();
+			status.setStatusCode(StatusCode.FAILED);
+			replacingRelease.getInfo().setStatus(status);
+			replacingRelease.getInfo().setDescription("Failed upgrade. Exception = [" +
+					e.getClass().getName() + "], Message = [" + e.getMessage() + "]");
+			this.releaseRepository.save(replacingRelease);
+		}
+		return null;
```
jvalkeal: U need to actually return `CompletableFuture` which is then completed with `complete(T)` or `completeExceptionally(Throwable)` instead of returning `null`.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/SimpleRedBlackUpgradeStrategy.java
```diff
@@ -108,50 +109,16 @@ public Release upgrade(Release existingRelease, Release replacingRelease,
...
+						replacingRelease.getName(), replacingRelease.getVersion());
 
-		// Update Status in DB
-		Status status = new Status();
-		status.setStatusCode(StatusCode.DEPLOYED);
-		replacingRelease.getInfo().setStatus(status);
-		replacingRelease.getInfo().setDescription("Upgrade complete");
+		// Executes on another thread.
+		// TODO Investigate why lookups in appDeployerDataRepository do not return expected results
+		this.healthCheckAndDeleteStep.waitForNewAppsToDeploy(existingRelease, existingAppDeployerData,
```
jvalkeal: Don't understand why this doesn't error in spring as I'd assume spring `@Async` should use `CompletableFuture` but this function just returns null. Maybe error is discarded or spring checks that actual returned value is null. Anyway this is somewhat wrong.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
jvalkeal: I think actual stuff what this method do should be fully wrapped in `Future` which is then executed when future value is requested.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/SimpleRedBlackUpgradeStrategy.java
jvalkeal: Also I'm not sure if whole `Transactional` stuff work anymore with this.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
```diff
@@ -172,9 +188,51 @@ public AppDeployerReleaseManager appDeployerReleaseManager(ReleaseRepository rel
...
+	}
+
+	@Bean(name = "skipperThreadPoolTaskExecutor")
+	@Override
+	public Executor getAsyncExecutor() {
+		return new ThreadPoolTaskExecutor();
+	}
+
+	@Override
+	public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
```
jvalkeal: Right ok, this might explain where execution errors end up what I mentioned in other comments.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/DeleteStep.java
markpollack: Right.  Will change to that if we fail to get the health within the timeout and in the general case of failure, we delete the apps we asked to deploy for the new release.  Status will be set to failed.  Will prob then need to change the 'findLatestRelease' to be findLatestReleaseNotFailedNotUnknown...

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HealthCheckAndDeleteStep.java
markpollack: I'll change this to void, it ended up being a `CompletableFuture` because I made some mistakes not having the method in a separate class and was spinning my wheels for a bit.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/SimpleRedBlackUpgradeStrategy.java
markpollack: Will look to split the logic into two transactions, such that there won't be two concurrent transactions occuring in the same request lifecycle.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
markpollack: In my testing, this never got called.  That is why I put a large catch block around all of the code in the `@Async` method.  Not sure what I am doing wrong.

ilayaperumalg:
The changes discussed in this PR were pushed as 1e7f58db7b4f0f039423bf4fc2fe0edc730b2140

PR #254, opened 2017-10-24
Handle null platform status values

### LINKED ISSUE #253: Upgrade operation response issue at the client side
Given the upgrade operation could return the null platform status, the client side throws error when it is null:

```
skipper:>install --package-name log --package-version 1.0.0 --release-name testing
Released testing
skipper:>upgrade --package-name log --package-version 2.0.0
org.springframework.shell.ParameterMissingResolutionException: Parameter '--release-name string' should be specified
Details of the error have been omitted. You can use the stacktrace command to print the full stacktrace.
```

ilayaperumalg:
- When the upgrade/rollback operations return null platform status (due to the async nature of the method), we need to handle it appropriately
  - Current change shows `unknown` status for both Release and Platform

Resolves #253


jvalkeal:
not really happy that we even need to do these changes but for now better than nothing. merging...

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-skipper/commit/597d7565cabe108ca4d7a970a9ad727e9ca219ea

PR #257, opened 2017-10-24
Use findByNameRequired for packagemetadata by name

### LINKED ISSUE #256: Use custom repo method for getting package metadata by name
Use `findByNameRequired` and throw SkipperException when there is no package metadata found with the given name.

ilayaperumalg:
- When packageMetadata is queried by the given name, we always throw SkipperException
and hence make this a custom repository method `findByNameRequired`
 - Add test

Resolves #256


jvalkeal:
lgtm

PR #272, opened 2017-10-25
Fail release if no upgrade performed

### LINKED ISSUE #268: Update operation should set the release status to failed when there is no difference
If the replacing release is found to be of no different from the existing release, then the replacing release should be marked `failed`.

ilayaperumalg:
- When there is no upgrade based on the ReleaseDifference, then the corresponding replacing release should be set to `FAILED` status.
 - Also, fail the upgrade before setting the manifest if there is no difference found

Resolves #268

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
@@ -163,6 +163,15 @@ public class AppDeployerReleaseManager implements ReleaseManager {
 	@Override
 	public ReleaseAnalysisReport createReport(Release existingRelease, Release replacingRelease) {
 		ReleaseAnalysisReport releaseAnalysisReport = this.releaseAnalyzer.analyze(existingRelease, replacingRelease);
+		if (releaseAnalysisReport.getReleaseDifference().areEqual()) {
+			Status status = new Status();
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
@jvalkeal Please verify this PR for `upgrade` for now. I will have another PR that fixes the rollback issue that will calculate the versions appropriately which will be on top of this PR.

jvalkeal:
I'm just questioning that we create a new release and save it to DB with FAILED state. That release has not exactly failed and say it's just a noise in a DB.

ilayaperumalg:
yeah, I agree. But, we don't seem to have any other possible state that could fit correctly here. Given, this operation is failed and we always persist this release object, `FAILED` turned out to be the interim choice :-)

jvalkeal:
And as we throw `SkipperException`, I'd expect any ongoing transaction to get rolled back. This doesn't happen which is a bit strange on its own.

ilayaperumalg:
> And as we throw SkipperException, I'd expect any ongoing transaction to get rolled back

We don't have to rollback and that's the main reason to set the status of the Release to be `FAILED` and apart from the creation of the Release in DB with `FAILED` stage, nothing happened in this transaction. Do you see any issue this way?

jvalkeal:
Right, looks like `org.springframework.cloud.skipper.server.service.ReleaseService.createReport(UpgradeRequest)` is called inhouse in a `RelaseService`, thus bypassing transaction. There should be transaction but then updating DB and throwing would be kinda pointless.

ilayaperumalg:
Closing this as we decided not to store failed release in DB in this scenario.

Issue #273, opened 2017-10-25
Move methods in ReleaseManager.upgrade into separate class to get Tx proxy functionality


markpollack:
Now we are calling the two release manager methods directly from the .upgrade method so the @Transactional does not kick in.

PR #274, opened 2017-10-25
Move the transactional method `createReport` into another class

markpollack:
* Move helper method for packageMetadata lookup into PackageMetadataRepository

Fixes GH-273

### DIFF
--- a/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/Info.java
+++ b/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/Info.java
@@ -84,4 +84,21 @@ public class Info extends AbstractEntity {
 	public void setDescription(String description) {
 		this.description = description;
 	}
+
+	/**
+	 * Create a new Info instance with the given description, first deployed and last deployed dates
+	 * set to the current date-time, and StatusCode.UNKNOWN.
+	 * @param description a string describing the general info status at a level finer than StatusCode
+	 * @return a new Info object
+	 */
+	public static Info createNewInfo(String description) {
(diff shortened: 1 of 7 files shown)


spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/Info.java
```diff
@@ -84,4 +84,21 @@ public String getDescription() {
...
 		this.description = description;
 	}
+
+	/**
+	 * Create a new Info instance with the given description, first deployed and last deployed dates
+	 * set to the current date-time, and StatusCode.UNKNOWN.
+	 * @param description a string describing the general info status at a level finer than StatusCode
+	 * @return a new Info object
+	 */
+	public static Info createNewInfo(String description) {
```
jvalkeal: just comment for future as we probably want to overhaul other pojos as well for static methods and think if we want to add Builder interfaces. Naming will easily get off the rails.

jvalkeal:
lgtm, hopefully merging in about 18mins.

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/Info.java
markpollack: Please create an issue to add the builder pattern here....

markpollack: So far, I think this is the only case, but we do have a few other sections with this type of 'boilerplate' code which could be moved down into the domain objects....

PR #281, opened 2017-10-26
Transaction handling during health check

### LINKED ISSUE #280: Transaction handling when doing the health check
When doing upgrade, the appDeployerData is persisted within the same transaction that the health check also uses to detect the health of the apps being deployed.

This makes the status check fail (with no appDeployerData for this release) until the health check is complete.

ilayaperumalg:
- Add @Transactional only to the health check method
 - Setting appdeployer data for th release being replaced doesn't belong to transaction as it is similar to the release report saving the replacing release

Resolves #280


jvalkeal:
I've been eyeballing this upgrade flow, possible errors and how transactions are handled. While I'm now starting to see multiple issues in a whole upgrade flow, don't think there's anything wrong in this PR other than it always raises a red flag when transaction is taken away from method interacting with repos.

markpollack:
Closing, will issue new PR based on discussions

PR #291, opened 2017-10-26
ReleaseStateUpdateService should be enabled by default

### LINKED ISSUE #290: ReleaseStateUpdateService should be enabled by default
Reverse the name of the property from `disable...` to `enable..` and add `matchIfMissing` to `@ConditionalOnProperty annotation`.

Also add `@EnabledScheduling` to make it clear that scheduling features are being used.

markpollack:
* Add local docs in three minute tour

Fixes #290


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
```diff
@@ -183,7 +185,7 @@ public ReleaseService releaseService(PackageMetadataRepository packageMetadataRe
 	}
 
 	@Bean
-	@ConditionalOnProperty("spring.cloud.skipper.server.disableReleaseStateUpdateService")
+	@ConditionalOnProperty(value = "spring.cloud.skipper.server.enableReleaseStateUpdateService", matchIfMissing = true)
```
ilayaperumalg: We could use `prefix`, something like: `@ConditionalOnProperty(prefix = "spring.cloud.skipper.server", name= "enableReleaseStateUpdateService",
			matchIfMissing = true)`

PR #323, opened 2017-11-11
Add initial documentation for using custom DB

AWeeks1:
Add instructions and guide for how to use custom DB (in this example postgres) instead of h2. 
[Didn't know where to add it, so made it an appendix, move wherever deemed relevant]

### DIFF
--- a/spring-cloud-skipper-docs/src/main/asciidoc/appendix-building.adoc
+++ b/spring-cloud-skipper-docs/src/main/asciidoc/appendix-building.adoc
@@ -55,3 +55,46 @@ You can generate eclipse project metadata using the following command:
 
 The generated eclipse projects can be imported by selecting `import existing projects`
 from the `file` menu.
+
+=== Using custom databse
+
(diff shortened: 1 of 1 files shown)


jvalkeal:
Maybe instead of documenting how to replace DB libs, package all supported DB's as default and then document settings. Even if you have h2/mysql/postgres/etc in a classpath, boot will back off using embedded db when you add settings to use external DB.

jvalkeal:
I'm merging this as is. There's going to be other changes to add full support for other DB's as well so we can revisit this section then.

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-skipper/commit/96d5b5071bf0d410d1dc8f7a79d41a31bde39d2d. Did polish asciidoc a little.

AWeeks1:
Sorry forgot to add here, for postgres there is a further change needed. This will work, but will crash when the commands `history` or `list` are used with error: 

```org.postgresql.util.PSQLException: Large Objects may not be used in auto-commit mode.```

To solve this, you need to add the @Transactional annotation to the `ReleaseRepository.java` in the Skipper-Server-Core.

PR #342, opened 2017-11-17
Tighten up `Package` section of documentation

AWeeks1:
Overall just made some small things clearer/more obvious. Some are style decisions so feel free to remove. 

Also, not as many substantive changes as I thought would be made. This section though, as someone who just had to use it to create a package, is not necessarily as clear or helpful as it could be. But I don't know if it's because of this section (which is pretty clear, though awkward to have to flick to 19 to know what the various fields mean), or because there is a missing section along the lines of "Making your own package" which walks through it more. Maybe that sort of customization section might be useful?


AWeeks1:
Sorry don't know how to undo a PR, keep only pushing single file changes instead of combining them all into one PR (sorry for that).

AWeeks1:
Closing: Included in https://github.com/spring-cloud/spring-cloud-skipper/pull/343

PR #354, opened 2017-11-20
Change dependencies leading to shell

### LINKED ISSUE #353: Change shell dependencies
Starting a shell now warns:
```
WARN : org.springframework.boot.context.properties.ConfigurationPropertiesBindingPostProcessor : The @ConfigurationProperties bean class org.springframework.cloud.deployer.spi.local.LocalDeployerProperties contains validation constraints but had not been annotated with @Validated.
```

While this could get fixed in a source itself, it feels a bit wrong for shell to get all deps defined in `spring-cloud-skipper` itself. This might have been leftover when some classes were organised in different packages, but we didn't change deps. Looks like all those deployer deps in `spring-cloud-skipper` are more or less meant for server, not shell.

jvalkeal:
- Fix deps so that client/shell doesn't get any deps
  it doesn't need.
- For spring-cloud-skipper, remove all deployer deps and just
  use spring-cloud-deployer-spi.
- For spring-cloud-skipper-client, bring back spring-web and httpclient.
  Previously httpclient were transitive deps via aether and spring-web via
  spring-cloud-deployer-local.
- Fixes #353

### DIFF
--- a/spring-cloud-skipper-client/pom.xml
+++ b/spring-cloud-skipper-client/pom.xml
@@ -15,6 +15,17 @@
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

PR #355, opened 2017-11-20
Update custom DB config doc

ilayaperumalg:


spring-cloud-skipper-docs/src/main/asciidoc/appendix-building.adoc
```diff
@@ -56,44 +56,22 @@ You can generate eclipse project metadata using the following command:
...
-```
-Which will build the jar, without running tests, creating documentation or checkstyle.
-
-Run `java -jar spring-cloud-skipper-server/target/spring-cloud-skipper-server-1.0.0.BUILD-SNAPSHOT.jar` if you updated the application.yml. If you created your own config, point to it by using:
-`java -jar spring-cloud-skipper-server/target/spring-cloud-skipper-server-1.0.0.BUILD-SNAPSHOT.jar --spring.config.location=/my-path/properties.ym;`. All the requests/history will now be stored on your custom DB!
-
-Important note, `spring.jpa.hibernate.ddl-auto` property will probably want to be changed to fit your use case. There are five options: `create`, `create-drop`, `update`, `validate` and `none`. `Create` and `create-drop` will delete all existing data in the DB on server boot-up (`create-drop` does it on server boot-down). So if you are using your own DB to persist state use `update`, `validate`, or `none`.
+=== Using a Custom Database
+
+Skipper, by default, uses an H2 DB that they control. It is possible to change the DB source, by specifying the
```
markpollack: 'they' is a funny way of saying it.  `Skipper, by default, uses an embedded H2 database is no `spring.datasource` property is specified.  ..

markpollack:
We should move this out of the building section of the docs, to somewhere in the getting started area.  It was in building since before the only way to change it was by a code change.

ilayaperumalg:
This was merged as 492c0d1059bcd83245adaf77a371303670ce191f

PR #367, opened 2017-11-28
Add statemachine for release handling

### LINKED ISSUE #234: Add statemachine for release handling
This can be considered as initial POC based on some existing designs.

jvalkeal:
- Uses 1.2.8 snapshot as some enhacement were required
  around persistense and utility classes.
- Machine design as presented in a ticket #234
- Minimal changes outside of new machine classes.
- Machine is only called from SkipperController which
  delegates to SkipperStateMachineService hiding low level
  logic.
- UpgradeStrategy still has its old method `upgrade` which
  has been split to four doing same upgrade, but in steps
  order for machine have better control over upgrade logic.
  Old `upgrade` method and its use(might still be used in tests)
  is removed later and it causes quite a bit of changes.
- Machine is executed in its own thread pool, thus machine logic
  is asynchronous. Blocking happens in SkipperController as
  `Release` is returned to caller.
- Fixes #234

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
@@ -30,6 +30,7 @@ import org.springframework.cloud.skipper.domain.UpgradeRequest;
 import org.springframework.cloud.skipper.domain.UploadRequest;
 import org.springframework.cloud.skipper.server.service.PackageService;
 import org.springframework.cloud.skipper.server.service.ReleaseService;
+import org.springframework.cloud.skipper.server.statemachine.SkipperStateMachineService;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ExceptionHandler;
@@ -62,6 +63,9 @@ public class SkipperController {
 	private String appVersion;
 
 	@Autowired
+	private SkipperStateMachineService stateMachineService;
+
+	@Autowired
 	public SkipperController(ReleaseService releaseService, PackageService packageService) {
 		this.releaseService = releaseService;
 		this.packageService = packageService;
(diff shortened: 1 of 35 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
```diff
@@ -61,6 +62,9 @@
 	@Value("${info.app.version:#{null}}")
 	private String appVersion;
 
+	@Autowired
+	private SkipperStateMachineService stateMachineService;
```
markpollack: Shouldn't have any autowiring and we create the controller in the SkipperServerConfiguration class.  Will modify on merge.

markpollack: It feels a bit awkward now that the Controller calls both `SkipperStateMachineService` and `ReleaseService` since  `SkipperStateMachineService` also calls `ReleaseService`.   Taking the `ReleaseService` methods ` list, history, manifest, status` and putting them in a new interface would make these relationships more direct in that `ReleaseService` only exposes the methods that are "orchestrated" by `SkipperStateMachineService`

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/statemachine/UpgradeStartAction.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	@Override
+	protected void executeInternal(StateContext<SkipperStates, SkipperEvents> context) {
+		setUpgradeCutOffTime(context);
+
+		UpgradeRequest upgradeRequest = context.getMessageHeaders().get(SkipperEventHeaders.UPGRADE_REQUEST, UpgradeRequest.class);
+		log.info("upgradeRequest {}", upgradeRequest);
+		if (upgradeRequest != null) {
+			ReleaseAnalysisReport releaseAnalysisReport = this.releaseReportService.createReport(upgradeRequest);
```
markpollack: This will throw an exception if we are asking to upgrade to a package that does not exist.  Eventually

```
org.springframework.cloud.skipper.SkipperException: Can not find package 'random', version '1.0.0'
	at org.springframework.cloud.skipper.server.repository.PackageMetadataRepositoryImpl.findByNameAndOptionalVersionRequired(PackageMetadataRepositoryImpl.java:86)
```
will get called.   However, the state machine does not stop executing the upgrade process and enters the state `UPGRADE_DEPLOY_TARGET_APPS`.  In which case it fails with an NPE since the `ReleaseAnalysis` object is null.

I am not yet familiar enough to make a change in the setup of the state machine to take this into account.  I added a null check and changed the test assertion in `ReleaseServiceTests.testInstallAndUpdatePackageNotFound` to match a new string, but we need to revisit this section of the code. 

I know you wanted to remove the `upgrade` method in another pass, but I wanted to get a bit more test coverage by exercising the statemachine

markpollack:
Merged as b79ddd0adce32b8ca70fc8747e1313e0fdf2c902

Removed the upgrade method in UpgradeStrategy.  

Need to fix handing of exceptions thrown in `UpgradeStartAction`.

PR #399, opened 2018-01-04
Refactor ReleaseAnalyzer

### LINKED ISSUE #384: Refactor ReleaseAnalyzer to keep data structure with differences between releases.
The current implementation builds a String that indicates there is a difference between one release an another.   There should instead be a simple boolean check based on the results of storing the different MapDifference instances and a new method created to "pretty print" what has changed.  This may also tie into the general 'show a diff' between two versions, though often the diff needs to be down in a pretty print style as well as a more version-control diff style (think git patches)

markpollack:
* Keep data structure with differences between releases.
* Provide helper classes to pretty print difference
* Add two deployer specific domain classes so clients can retrieve the diffs

Fixes #384


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ApplicationManifestDifferenceFactory.java
```diff
@@ -0,0 +1,132 @@
...
+				getResourceAndVersionDifference(existingApplicationManifest, replacingApplicationManifest),
+				getApplicationPropertiesDifference(existingApplicationManifest, replacingApplicationManifest),
+				getDeploymentPropertiesDifference(existingApplicationManifest, replacingApplicationManifest));
+	}
+
+	protected MapDifference getApiAndKindDifference(
+			SpringCloudDeployerApplicationManifest existingApplicationManifest,
+			SpringCloudDeployerApplicationManifest replacingApplicationManifest) {
+		Map<String, String> existingApiAndKindMap = new HashMap<>();
+		existingApiAndKindMap.put("api", existingApplicationManifest.getApiVersion().trim());
```
ilayaperumalg: Isn't this `apiVersion` instead of `api`? I think we need to use the constants defined in `SpringCloudDeployerApplicationManifest` instead of String values for the keys.

ilayaperumalg: ok, I see that as long as we use the same key `api` in the other comparing (replacingApplicationManifest) map we are good. But, we can still use `apiVersion` for consistency I guess.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ApplicationManifestDifferenceSummaryGenerator.java
```diff
@@ -0,0 +1,104 @@
...
+	/**
+	 * Generate a textual summary of the ApplicationManifestDifference
+	 * @param applicationManifestDifference the ApplicationManifestDifference
+	 * @return a textual summary
+	 */
+	public String generateSummary(ApplicationManifestDifference applicationManifestDifference) {
+		StringBuffer stringBuffer = new StringBuffer();
+
+		if (applicationManifestDifference.areEqual()) {
+			stringBuffer.append("Existing and Replacing Application Name=["
```
ilayaperumalg: Application -> Application(s) and can possibly use `Name` at the end something like:

`Existing and Replacing Applications are equal for the applications named <>`

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
```diff
@@ -101,45 +100,57 @@ private ReleaseAnalysisReport analyzeDependentPackagesOnly(
...
+	}
+
+	private ReleaseAnalysisReport createReleaseAnalysisReport(Release existingRelease,
+			Release replacingRelease,
+			List<String> appsToDelete,
+			List<ApplicationManifestDifference> applicationManifestDifferences) {
+		ReleaseDifference releaseDifference = new ReleaseDifference();
+		releaseDifference.setDifferences(applicationManifestDifferences);
+		if (!releaseDifference.areEqual()) {
+			logger.info("Differences detected between existing and application manifests. Upgrading applications = [" +
```
ilayaperumalg: missing `replacing`

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
```diff
@@ -101,45 +100,57 @@ private ReleaseAnalysisReport analyzeDependentPackagesOnly(
...
+		List<ApplicationManifestDifference> applicationManifestDifferences = new ArrayList<>();
+
+		ApplicationManifestDifference applicationManifestDifference = applicationManifestDifferenceFactory
+				.createApplicationManifestDifference(
+						existingApplicationSpecList.get(0).getApplicationName(),
+						existingApplicationSpecList.get(0),
+						replacingApplicationSpecList.get(0));
+		applicationManifestDifferences.add(applicationManifestDifference);
+
+		return createReleaseAnalysisReport(existingRelease, replacingRelease, appsToDelete,
```
ilayaperumalg: I think we can get rid of `appsToDelete` (also, this should have `appsToUpdate` I believe) local variable in these `analyze` methods as the appsToUpdate are now retrieved from the difference directly when creating the analysis report. Hence, this `appsToDelete` parameter can be removed as well.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
```diff
@@ -101,45 +100,57 @@ private ReleaseAnalysisReport analyzeDependentPackagesOnly(
...
+				applicationManifestDifferences);
+	}
+
+	private ReleaseAnalysisReport createReleaseAnalysisReport(Release existingRelease,
+			Release replacingRelease,
+			List<String> appsToDelete,
```
ilayaperumalg: Can better change `appsToDelete` -> something like  `appsToUpdate`

spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/statemachine/StateMachineTests.java
```diff
@@ -134,32 +133,32 @@ public void testSimpleInstallShouldNotError() throws Exception {
 		Mockito.when(releaseService.install(any(), any())).thenReturn(new Release());
 
 		Message<SkipperEvents> message = MessageBuilder
-			.withPayload(SkipperEvents.INSTALL)
-			.setHeader(SkipperEventHeaders.PACKAGE_METADATA, new PackageMetadata())
-			.setHeader(SkipperEventHeaders.INSTALL_PROPERTIES, new InstallProperties())
-			.setHeader(SkipperEventHeaders.VERSION, 1)
-			.build();
+				.withPayload(SkipperEvents.INSTALL)
```
ilayaperumalg: Is there a code formatter update we can do to maintain the same format (especially the statemachine related builder code)

jvalkeal: In a same way than reactor stuff, these are manually formatted for things to stay readable.

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/deployer/ApplicationManifestDifference.java
```diff
@@ -0,0 +1,99 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.MapDifference;
+
+import org.springframework.util.Assert;
+
+/**
+ * Describes the difference between two ApplicationManifests
+ * @author Mark Pollack
+ */
+public class ApplicationManifestDifference {
```
jvalkeal: Why to move these from server to a shared package? Are we planning to use these in a client?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalysisReport.java
```diff
@@ -51,25 +52,39 @@ public ReleaseAnalysisReport(List<String> applicationNamesToUpgrade, ReleaseDiff
...
-		return applicationNamesToUpgrade;
+		return this.applicationNamesToUpgrade;
 	}
 
 	public ReleaseDifference getReleaseDifference() {
-		return releaseDifference;
+		return this.releaseDifference;
+	}
+
+	public String getReleaseDifferenceSummary() {
```
jvalkeal: Maybe `ReleaseDifferenceSummaryGenerator` should generate all this and not just modify what it returns.

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/deployer/ApplicationManifestDifference.java
```diff
@@ -0,0 +1,99 @@
...
+		this.resourceAndVersionDifference = resourceAndVersionDifference;
+		this.applicationPropertiesDifference = applicationPropertiesDifference;
+		this.deploymentPropertiesDifference = deploymentPropertiesDifference;
+	}
+
+	public String getApplicationName() {
+		return this.applicationName;
+	}
+
+	public MapDifference getApiAndKindDifference() {
```
jvalkeal: If this is somehow exposed via rest as json, what's the actual format? I know that next step would be to remove guava but I'm slightly worried what we expect as json.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ApplicationManifestDifferenceFactory.java
markpollack: Sure, will change to `apiVersion`

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ApplicationManifestDifferenceSummaryGenerator.java
markpollack: Do you mean `applications named <time>` or `applications named time`.  I ended up with the `Name[time]` approach since otherwise it didn't seem to stand out when looking at the reports.  I added a few println's in the tests that I later removed....

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
markpollack: Will fix

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
markpollack: I will investigate.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
markpollack: Yes, also depends on the results of the previous comment if it should stick around at all.

spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/statemachine/StateMachineTests.java
markpollack: yea, I got a bit nervous when I hit reformat there, I prob needed to just optimize the imports.  There is an option I believe to turn off the code formatter for certain sections (we do this in the shell code).  I'll look to revert to the old formatted code and add those conditional code formatting guards.

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/deployer/ApplicationManifestDifference.java
markpollack: I wanted to address that in another PR, but yes, I see them moving into the client package, but not in the domain package.

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/deployer/ApplicationManifestDifference.java
markpollack: I remember looking at the fields in MapDifference once and thinking it was going to marshall/unmarshall to json well....  let's have a look.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ApplicationManifestDifferenceSummaryGenerator.java
markpollack: Changed to
`Existing and Replacing Applications are equal for Application Name=[time]`
and
`Existing and Replacing are different for Application Name=[time]`

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
markpollack: I see your point now after looking at the code.  I will move the creation of the list into the utility method `createReleaseAnalysisReport` (and rename it to appsToUpdate) but it will remain as a property of the `ReleaseAnalysisReport`.

spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/statemachine/StateMachineTests.java
markpollack: Can't seem to get `//CHECKSTYLE:OFF` working so just reverting back to the manually formatted code.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalysisReport.java
markpollack: I'll pass in as arguments the existingRelease and replacing release's name/version to encapsulate this better.  Passing in the whole release object seems like a bit overkill...

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/deployer/ApplicationManifestDifference.java
markpollack: It has Maps as private fields and a ValueDifference object, which is a simple pair of primitives.  I haven't tested but we should be able to make an equivalent version of this that ser/deser's correctly over JSON.

jvalkeal:
lgtm merging

jvalkeal:
merged/squashed per https://github.com/spring-cloud/spring-cloud-skipper/commit/08daf78052aa66bde9c6a4fb17545195175d6fcc

PR #404, opened 2018-01-08
Add custom properties diff

### LINKED ISSUE #377: Remove guava from the stack
As a developer, I'd want to remove guava from the stack.

jvalkeal:
- Remove guava and replace its properties difference
  functionality with a custom one.
- New implementation is PropertiesDiff and contains features
  for left and right to see added, removed, common and
  changed properties.
- Fixes #377

### DIFF
--- a/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/support/PropertiesDiff.java
+++ b/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/support/PropertiesDiff.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/support/PropertiesDiff.java
```diff
@@ -0,0 +1,228 @@
...
+import java.util.Map.Entry;
+
+import org.apache.commons.lang3.builder.Diff;
+import org.apache.commons.lang3.builder.DiffBuilder;
+import org.apache.commons.lang3.builder.DiffResult;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+/**
+ * {@code PropertiesDiff} is an implementation to come up with a difference
+ * between two properties sets. This difference gives answer to questions like
```
markpollack: between two string based maps.

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/support/PropertiesDiff.java
```diff
@@ -0,0 +1,228 @@
...
+		this.common = common;
+		this.changed = changed;
+	}
+
+	/**
+	 * Gets the removed properties.
+	 *
+	 * @return the removed
+	 */
+	public Map<String, String> getRemoved() {
```
markpollack: I'll merge, but interested to know if you think a name more along the lines of 'entriesOnlyOnLeft` would be more consistent since the builder uses 'left' and 'right' methods, this way the terminology stays the same.

markpollack:
Made small javadoc change, merged as a7f05a9b10e83bc57088910a93cd8822ccaa900c

Let's discuss the naming in another issue.

PR #410, opened 2018-01-09
Change SkipperClientProperties uri property to serverUri

### LINKED ISSUE #368: Rename "uri" to "serverUri" in SkipperClientProperties
The `host` property should just be the hostname.  Now it also includes `/api`, eg. `http://localhost:8080/api`. 

Perhaps the correct fix is to introduce a new property named `skipper.client.uri`.

Also, should we have the `spring.cloud` prefix as well?

markpollack:
Fixes #368


ilayaperumalg:
LGTM, merging.

PR #411, opened 2018-01-09
Use spec.version in generic template

### LINKED ISSUE #407: In generic-template.yml change use of {{version}} to {{spec.version}}
This would be more consistent,  though not as convenient to type

From

```
upgrade --release-name ticktockskipper --package-name ticktock --properties log.version=1.1.1.RELEASE
```

To
```
upgrade --release-name ticktockskipper --package-name ticktock --properties log.spec.version=1.1.1.RELEASE
```

It prob also helps to limit the number of 'special' keywords to just `spec` and `metadata` wrt to handling the properties to yml conversion

ilayaperumalg:
- Update the template and doc files

Resolves #407

### DIFF
--- a/spring-cloud-skipper-docs/src/main/asciidoc/three-hour-tour.adoc
+++ b/spring-cloud-skipper-docs/src/main/asciidoc/three-hour-tour.adoc
@@ -247,8 +247,8 @@ metadata:
   type: sample
 spec:
   resource: maven://org.mysample:mypackage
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
I believe this change doesn't require changes in the test packages' templates as they use values.yml to inject the `version` property.

tzolov:
Rebased and merged at c90900e

PR #413, opened 2018-01-09
Redacts sensitive info from manifest info in debug messages.

### LINKED ISSUE #229: Mask sensitive data from being logged
There are places (especially in exception handlers), we seem to log manifest information etc.,

Since this could be critical information not intended to be logged, we can remove and replace it with the appropriate error messages.

cppwfs:
This PR redacts 2 debug messages that dump the YAML to the log.  Are there other spots that this PR doesn't cover?

resolves #229


ilayaperumalg:
@cppwfs Looks great! I will do some checks (also, if there are any other places this change requires update) and merge it.

ilayaperumalg:
LGTM, merging.

PR #435, opened 2018-01-12
Package (name version repoId) must be unique

### LINKED ISSUE #317: Skipper Shell: Uploading a package with same name and version that already exists should throw error
Currently, using the `upload --path` command you can upload a package with the same `package-name` and `package-version` as one that already exists. Can cause some problems, mostly seems to default to first manifest given (working on proving this). 

Would be nice if `upload` threw an error if it gave an identical package already existed. 

Adding code to here:
https://github.com/spring-cloud/spring-cloud-skipper/blob/master/spring-cloud-skipper-shell-commands/src/main/java/org/springframework/cloud/skipper/shell/command/SkipperCommands.java 

Haven't tried the code yet, but the code from 154-169 should be reusable in the try on 318.

tzolov:
- Package with same Name, Version and RepoId must be unique

Resolves #317

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/PackageService.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/PackageService.java
@@ -31,11 +31,11 @@ import org.slf4j.LoggerFactory;
 import org.zeroturnaround.zip.ZipUtil;
 
 import org.springframework.cloud.skipper.SkipperException;
-import org.springframework.cloud.skipper.io.PackageFileUtils;
 import org.springframework.cloud.skipper.domain.Package;
 import org.springframework.cloud.skipper.domain.PackageMetadata;
 import org.springframework.cloud.skipper.domain.Repository;
 import org.springframework.cloud.skipper.domain.UploadRequest;
+import org.springframework.cloud.skipper.io.PackageFileUtils;
 import org.springframework.cloud.skipper.io.PackageReader;
 import org.springframework.cloud.skipper.io.TempFileUtils;
 import org.springframework.cloud.skipper.server.repository.PackageMetadataRepository;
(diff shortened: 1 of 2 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/PackageService.java
```diff
@@ -260,6 +260,17 @@ private void validateUploadRequest(UploadRequest uploadRequest) {
 				+ uploadRequest.getExtension());
 		Assert.notNull(uploadRequest.getPackageFileAsBytes(), "Package file as bytes must not be null");
 		Assert.isTrue(uploadRequest.getPackageFileAsBytes().length != 0, "Package file as bytes must not be empty");
+
+		Repository repository = this.repositoryRepository.findByName(uploadRequest.getRepoName());
+		if (repository != null) {
+			PackageMetadata existingPackageMetadata = this.packageMetadataRepository.findByRepositoryIdAndNameAndVersion(
```
markpollack: I've created #436 so that we can look up directly by name.

spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/service/PackageServiceTests.java
```diff
@@ -154,6 +155,14 @@ public void upload() throws Exception {
 		assertThat(downloadedPackage.getTemplates()).isNotNull();
 		assertThat(downloadedPackage.getConfigValues()).isNotNull();
 
+		try {
+			this.packageService.upload(uploadProperties);
+			fail("Attempt to upload a package with same name, version withing the same repository must fail!");
+		}
+		catch (SkipperException se) {
+			assertThat(se.getMessage()).contains("Package [log:9.9.9] already exists");
```
markpollack: Add repo name to error message

markpollack:
Minor fixes, merged as 81d1d4128d6ac99f164bdcecca0731111da8a709

PR #447, opened 2018-01-14
Add package delete command

### LINKED ISSUE #331: Create package delete command
Currently, there is no way to delete or remove a package that has been uploaded. This can cause problems if you are only changing some metadata name + version (See https://github.com/spring-cloud/spring-cloud-skipper/issues/317) or if you want to make a small change to a package for testing (or if you upload a package by mistake) etc etc. 

This can be fixed with an opposite method to ` upload --path` which could be `remove package --package-name --package-version`, which would remove the package instantly and update the local db.

markpollack:
* Only deletes if the latest releases using it are in the DELETED state.
* Only applies to local repositories.
* Updated RESTDocs

Fixes #331

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/RootController.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/RootController.java
@@ -29,7 +29,7 @@ import org.springframework.web.servlet.view.RedirectView;
 public class RootController {
 
 	/**
-	 * Handles the root URL of Skipper. Redirects users to the REST API entry point at
+	 * Handles the root URL of Skipper. Redipacrects users to the REST API entry point at
 	 * {@code /api }.
 	 *
 	 * @return RedirectView to api
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
@@ -26,6 +26,7 @@ import org.springframework.cloud.skipper.domain.PackageMetadata;
 import org.springframework.cloud.skipper.domain.Release;
 import org.springframework.cloud.skipper.domain.UpgradeRequest;
 import org.springframework.cloud.skipper.domain.UploadRequest;
+import org.springframework.cloud.skipper.server.service.PackageMetadataService;
 import org.springframework.cloud.skipper.server.service.PackageService;
 import org.springframework.cloud.skipper.server.service.ReleaseService;
 import org.springframework.cloud.skipper.server.statemachine.SkipperStateMachineService;
@@ -54,6 +55,8 @@ public class SkipperController {
 
 	private final PackageService packageService;
 
+	private final PackageMetadataService packageMetadataService;
+
 	@Value("${info.app.name:#{null}}")
 	private String appName;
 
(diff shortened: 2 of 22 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/PackageService.java
```diff
@@ -77,6 +77,11 @@ public PackageService(RepositoryRepository repositoryRepository,
 		this.packageReader = packageReader;
 	}
 
+	@Transactional
```
markpollack: This should be deleted.  Moved the functionality into PackageMetadataService as it had more in common with the collaborators in that class.

markpollack:
Note this comment in the code

```
 +	//TODO don't expose the ID, use names instead.
 +
 +	private Long packageMetadataId;
 +
 +	private Long repositoryId;
```

Would require some other reworking, can consider after this PR is merged.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
```diff
@@ -63,18 +66,29 @@
...
+
 	@RequestMapping(path = "/upload", method = RequestMethod.POST)
 	@ResponseStatus(HttpStatus.CREATED)
 	public PackageMetadata upload(@RequestBody UploadRequest uploadRequest) {
 		return this.packageService.upload(uploadRequest);
 	}
 
+	@RequestMapping(path = "/package/{name}", method = RequestMethod.DELETE)
+	@ResponseStatus(HttpStatus.OK)
+	public void packageDelete(@PathVariable("name") String name) {
```
ilayaperumalg: While this REST endpoint allows us to delete all the package metadata by the given name, delete by name/version could definitely be more useful. We can optionally allow specifying any valid `local` repo name to search for the package metadata as well.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/PackageMetadataService.java
```diff
@@ -50,10 +58,83 @@
...
+		List<String> activeReleaseNames = new ArrayList<>();
+		for (Release release : latestReleaseMap.values()) {
+			if (!release.getInfo().getStatus().getStatusCode().equals(StatusCode.DELETED)) {
+				canDelete = false;
+				activeReleaseNames.add(release.getName());
+			}
+		}
+		if (!canDelete) {
+			Repository repository = this.repositoryRepository.findOne(packageMetadata.getRepositoryId());
+			throw new SkipperException(String.format("Can not delete Package Metadata [%s:%s] in Repository [%s]. " +
```
ilayaperumalg: Given we iterate over the list of all the package metadata by the given `name`, throwing exception for the specific package metadata here (when the condition is met) would break the execution for the rest.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/SkipperController.java
markpollack: Yes, but we can add additional variations another time

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/PackageMetadataService.java
markpollack: I don't understand the comment.  

The method argument takes a specific packageMetadata.  We iterate overall the releases in local repositories and see if any have a status different than DELETED.  If so, then I think we can just abort any attempt to delete anything, until all releases are in the DELETED state.

ilayaperumalg: I meant the iteration done [here](https://github.com/spring-cloud/spring-cloud-skipper/pull/447/files/8ad87dc11106938a5cc130c0b6d3838ff4db632b#diff-2aa49f674c378b62d111b7898ed5526aR77) which subsequently calls this method.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/RootController.java
```diff
@@ -29,7 +29,7 @@
 public class RootController {
 
 	/**
-	 * Handles the root URL of Skipper. Redirects users to the REST API entry point at
+	 * Handles the root URL of Skipper. Redipacrects users to the REST API entry point at
```
ilayaperumalg: will fix while merging.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 5acdd08 with the couple of changes (cleanup unused code and typo fix) mentioned in the PR.

PR #472, opened 2018-01-17
Change controllers to return Resource<T>

### LINKED ISSUE #468: Package and ReleaseController to return Resource<T> objects
update DefaultSkipperClient to match

markpollack:
* Removed history method with max results, was odd error in jackson deser.  Can introduce later.

Fixes #468


markpollack:
need to rebase

PR #475, opened 2018-01-17
Update DELETE request

### LINKED ISSUE #444: Delete functionality in controllers should use RequestMethod.DELETE
Some are using RequestMethod.POST

ilayaperumalg:
- Use DELETE HTTP request instead of POST
 - Update ReleaseController `delete` method
   - Since this causes removal of request body in DELETE request, update the REST endpoint to use `deletePackage` as a path variable
   - Add a default DELETE request method that has `deletePackage` set to `false` in DeleteProperties
 - Update security config for the DELETE request endpoint changes
 - Update tests

Resolves #444

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
@@ -90,7 +90,7 @@ public class ReleaseController {
 		resource.add(
 				ControllerLinkBuilder.linkTo(methodOn(ReleaseController.class).rollback(null, 123))
 						.withRel("rollback"));
-		resource.add(ControllerLinkBuilder.linkTo(methodOn(ReleaseController.class).delete(null, null))
+		resource.add(ControllerLinkBuilder.linkTo(methodOn(ReleaseController.class).delete(null, true))
 							.withRel("delete"));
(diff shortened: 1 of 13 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
```diff
@@ -141,10 +141,19 @@ public ReleaseControllerLinksResource resourceLinks() {
 		return this.releaseResourceAssembler.toResource(release);
 	}
 
-	@RequestMapping(path = "/delete/{name}", method = RequestMethod.POST)
-	@ResponseStatus(HttpStatus.CREATED)
+	@RequestMapping(path = "/{name}", method = RequestMethod.DELETE)
+	@ResponseStatus(HttpStatus.OK)
+	public Resource<Release> delete(@PathVariable("name") String releaseName) {
+		Release release = this.skipperStateMachineService.deleteRelease(releaseName, new DeleteProperties());
```
tzolov: perhaps you should call the `ReleaseController#delete(releaseName, false)` instead of duplicating the implementation? Or perhaps factor out the implementation in private method and call it from both controller delete methods

markpollack:
to avoid a failed build in SCDF best to disable the CI for SCDF when this change is made to avoid an automatic trigger

tzolov:
Rebased and merged at 159db57

PR #476, opened 2018-01-17
Add Semantic versioning validation

### LINKED ISSUE #459: Add validation of package version against semver library
ATM we are relying on ascii ordering of the version and not constraining it to the semver spec leaves us open for this approach to fail.

Code needs to be added to `PackageService.validateUploadRequest`

Adding the validation will hopefully let us 'get away' using ascii desc ordering to get the latest version of a package.

There are two libraries we can use.

https://github.com/zafarkhaja/jsemver

which is widely used but dead

and

https://github.com/vdurmont/semver4j

which is not as widely used but looks active and maintained by netflix.

semver4j seems like the right choice.

ilayaperumalg:
- For package upload request, validate the version of the package using the Semantic versioning support library `https://github.com/zafarkhaja/jsemver`
 - Add tests

Resolves #459


tzolov:
I've added #483 and #484 to handle some additional edge cases.

tzolov:
Rebased and merged at d9aaea6

Issue #461, opened 2018-01-16
Prepare schemas for the release


sabbyanandan:
Review the schemas and prepare them with the "alter if not exist" / "create if not exist" clauses.

- [ ] MySQL - Vinny
- [ ] H2 - Vinny
- [ ] Db2 - Vinny
- [ ] HSQL - Vinny
- [ ] Postgres - Vinny
- [ ] SQLServer - Vinny 
- [ ] Oracle - Vinny

**Acceptance:**
Start the server; run a sample upload/install; restart the server; verify the table and the data remain undisturbed.

sabbyanandan:
Get rid of all the drops and replace it with "create if not exist".

sabbyanandan:
Delete "insert into hibernate_sequence values ( 1 );" entries.

sabbyanandan:
Move all the alter table commands towards the bottom.

cppwfs:
will need to merge https://github.com/spring-cloud/spring-cloud-skipper/pull/467 first, before working on this.

Issue #433, opened 2018-01-12
Remove generate-ddl:true from application.yml


markpollack:
Test that sql scripts are loading against h2 and mysql.

cppwfs:
I just ran a test where I renamed the postgresql94.sql to just postgresql.sql and boot used the correct sql file.  (Once I added the sql files to the resources).  I'll create a PR adding the sql files to the resources.

sabbyanandan:
So, once #467 is merged, we don't have to mess with `spring.datasource.platform` property then? 

If yes, the scope of this story simply the removal of the "generate-ddl" property.

Issue #432, opened 2018-01-12
Regenerate schema files based on #430 #425


markpollack:
If implementing lazy loading of package byte array, regenerate schemas.

PR #485, opened 2018-01-17
Add Flyway support for the DB schemas

viniciusccarvalho:
Moved scripts to db/migration/{vendor}
Tested on H2, HSQLDB, mysql, postgres, SQLServer
Oracle is having some issues when bootstraping flyway's schema_version table

Resolves spring-cloud/spring-cloud-skipper#471
Resolves spring-cloud/spring-cloud-skipper#461
Resolves spring-cloud/spring-cloud-skipper#433
Resolves spring-cloud/spring-cloud-skipper#432

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
         <spring-cloud-deployer-local.version>1.3.0.M2</spring-cloud-deployer-local.version>
         <reactor.version>3.0.7.RELEASE</reactor.version>
         <spring-shell.version>2.0.0.RELEASE</spring-shell.version>
-        <spring-statemachine.version>1.2.8.RC3</spring-statemachine.version>
+        <spring-statemachine.version>1.2.8.BUILD-SNAPSHOT</spring-statemachine.version>
         <jsonpath.version>2.2.0</jsonpath.version>
         <equalverifier.version>2.4.1</equalverifier.version>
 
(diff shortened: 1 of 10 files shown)


ghillert:
We should probably reference the Oracle dependencies via a Maven Profile, e.g.:

```
        <profile>
          <id>oracle</id>
          <repositories>
              <repository>
                <id>maven.oracle.com</id>
                <releases>
                  <enabled>true</enabled>
                </releases>
                <url>https://maven.oracle.com</url>
              </repository>
            </repositories>
            <dependencies>
              <dependency>
                <artifactId>ojdbc8</artifactId>
                <groupId>com.oracle.jdbc</groupId>
                <version>12.2.0.1</version>
              </dependency>
            </dependencies>
        </profile>
```

And add documentation to the reference guide on where to find information on how to enable the Oracle Maven repository. We have to be careful here not to violate the ASL. (Maybe double-check if the approach above would be doable, or whether we should not even reference the Oracle drivers via the pom.)

markpollack:
@ghillert How does it work with referencing oracle mvn coords in the pom - don't you need some credentials to download that driver?

markpollack:
removed pom ref to oracle driver, squashed and merged as 7841b564e81cc5689689e7a39b68961542ddaa10

PR #491, opened 2018-01-18
Polish controller error handling

### LINKED ISSUE #490: Polish controller error handling
Looks like when package is deleted and PackageDeleteException exists, it's not propagated via SkipperErrorAttributes. Also SkipperException is not handled at all.

jvalkeal:
- Instead of introducing global error handling, in PackageController
  and ReleaseController explicitely handle all skipper related
  exceptions so that we don't get 500's and SkipperErrorAttributes
  gets a change to kick in.
- As long as we throw ReleaseNotFoundException, PackageDeleteException or
  generic SkipperException, we don't get 500's and map to more meaninful
  code and reason.
- In SkipperErrorAttributes, override message set by boot's
  DefaultErrorAttributes to have one in our own skipper related exceptions.
  Otherwise we get message set in @ResponseStatus with @ExceptionHandler.
- Fixes #490

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/PackageController.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/PackageController.java
@@ -16,6 +16,8 @@
 package org.springframework.cloud.skipper.server.controller;
 
 import org.springframework.cloud.skipper.PackageDeleteException;
+import org.springframework.cloud.skipper.ReleaseNotFoundException;
+import org.springframework.cloud.skipper.SkipperException;
 import org.springframework.cloud.skipper.domain.InstallProperties;
 import org.springframework.cloud.skipper.domain.InstallRequest;
 import org.springframework.cloud.skipper.domain.PackageMetadata;
@@ -30,7 +32,6 @@ import org.springframework.hateoas.Resource;
 import org.springframework.hateoas.ResourceSupport;
 import org.springframework.hateoas.mvc.ControllerLinkBuilder;
 import org.springframework.http.HttpStatus;
-import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
@@ -38,9 +39,6 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
-
-import java.util.HashMap;
-import java.util.Map;
 
 import static org.springframework.hateoas.mvc.ControllerLinkBuilder.methodOn;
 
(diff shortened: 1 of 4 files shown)


jvalkeal:
Btw we now get better error messages through:

`http://localhost:7577/api/package/log`
```
{
    "timestamp": 1516286066140,
    "status": 412,
    "error": "Precondition Failed",
    "exception": "org.springframework.cloud.skipper.PackageDeleteException",
    "message": "Can not delete package [log], associated repository [test] is remote.",
    "path": "/api/package/log"
}
```

`http://localhost:7577/api/package/mylog`
```
{
    "timestamp": 1516286070568,
    "status": 409,
    "error": "Conflict",
    "exception": "org.springframework.cloud.skipper.SkipperException",
    "message": "Can not find a package named 'mylog'",
    "path": "/api/package/mylog"
}
```

`http://localhost:7577/api/package/install`
```
{
    "timestamp": 1516286131076,
    "status": 409,
    "error": "Conflict",
    "exception": "org.springframework.cloud.skipper.SkipperException",
    "message": "Release with the name [mylog] already exists and it is not deleted.",
    "path": "/api/package/install"
}
```


`http://localhost:7577/api/release/status/foo`
```
{
    "timestamp": 1516286145725,
    "status": 404,
    "error": "Not Found",
    "exception": "org.springframework.cloud.skipper.ReleaseNotFoundException",
    "message": "Release with the name [foo] doesn't exist",
    "path": "/api/release/status/foo",
    "releaseName": "foo"
}
```

ilayaperumalg:
Looks great!

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/PackageController.java
```diff
@@ -108,13 +106,22 @@ public void packageDelete(@PathVariable("name") String name) {
 		this.packageMetadataService.deleteIfAllReleasesDeleted(name, PackageMetadataService.DEFAULT_RELEASE_ACTIVITY_CHECK);
 	}
 
+	@ResponseStatus(value = HttpStatus.NOT_FOUND, reason = "Release not found")
```
ilayaperumalg: Maybe we can do some refactoring by having the same exception handlers in both the controllers. I guess this is fine for now.

jvalkeal: That would be the global error handling. Don't wanna go there for now as it's a bit more dark territory.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/PackageController.java
```diff
@@ -108,13 +106,22 @@ public void packageDelete(@PathVariable("name") String name) {
...
+	@ExceptionHandler(ReleaseNotFoundException.class)
+	public void handleReleaseNotFoundException() {
+		// needed for server not to log 500 errors
+	}
+
+	@ResponseStatus(value = HttpStatus.PRECONDITION_FAILED, reason = "Package deletion error")
```
tzolov: I think that `412 Precondition Failed` is not a correct message status.
 Although the message name sound relevant, HTTP has very concrete semantics related to conditional requests  such as  `If-Unmodified-Since` or `If-None-Match`. (https://tools.ietf.org/html/rfc7232#section-4.2)
IMO we should use HTTP 409 (CONFLICT) instead

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
```diff
@@ -181,13 +179,16 @@ public void handleReleaseNotFoundException() {
 		// needed for server not to log 500 errors
 	}
 
+	@ResponseStatus(value = HttpStatus.PRECONDITION_FAILED, reason = "Package deletion error")
```
tzolov: Same

ilayaperumalg:
Merging with the changes suggested.

ilayaperumalg:
Merged as bb046ce

PR #494, opened 2018-01-18
UploadRequest Name and Version must match those in package.yum

### LINKED ISSUE #483: Validate the package uploadRequest name and version match those in package.yum
Currently the client can create uploadRequest with a name and version different from those defined in the package.yml
The shell client for example uses the file name to extract the package name and version despite the name and version in the archive

tzolov:
- Add additional validation in the package service upload
 - Add test
Resolves #483


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as f852188

PR #540, opened 2018-01-31
Increase blob columns for DB2/HSQLDB.  Use @Lob for URIs in Repository

### LINKED ISSUE #538: Use @Lob for URI fields in Repository
255 is too small.

### LINKED ISSUE #539: Increase size of blob for storing package.zip file
Some blob fields are set to 255

markpollack:
Fixes #539
Fixes #538


ghillert:
Merged with commit 214fa100b256eb810eba8825268512df17b9eeb0

PR #567, opened 2018-03-16
Add starter and bom

### LINKED ISSUE #559: Add starter and bom
Using a normal model we need to have `spring-cloud-skipper-dependencies` and `spring-cloud-starter-skipper-server`. Using server in starter in case at some point we want to add starter for client.

### LINKED ISSUE #560: Fix autoconfig structure
Boot autoconfig is defined in a wrong place as `spring.factories` is in `spring-cloud-skipper-server` while its referenced class is in `spring-cloud-skipper-server-core`. Custom skipper app would not use `spring-cloud-skipper-server` which solely exists to create a fatjar as OOB packaged default impl.

jvalkeal:
- Add normal starter and bom concepts
- Combine all auto-configs into its own jar, better
  than 3 different jars having auto-configs where
  one defined auto-config class physicall existing in
  different jar.
- Clean how versions in poms are used.
- Fixes #559
- Fixes #560 

Comparing current snapshot with this PR, difference is:
```
diff /tmp/fatjar-current.out /tmp/fatjar-new.out
121a122
> BOOT-INF/lib/spring-cloud-skipper-autoconfigure-1.0.2.BUILD-SNAPSHOT.jar
```

There is custom skipper maven and gradle builds in https://github.com/jvalkeal/randomstuff/tree/master/custom-skipper-server

Fatjar diff against custom maven build against server master build using this PR
```
121a122
> BOOT-INF/lib/spring-cloud-starter-skipper-server-1.0.2.BUILD-SNAPSHOT.jar
```

Fatjar diff against custom gradle build against server master build using this PR
```
121a122
> BOOT-INF/lib/spring-cloud-starter-skipper-server-1.0.2.BUILD-SNAPSHOT.jar
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,9 @@
         <module>spring-cloud-skipper-docs</module>
         <module>spring-cloud-skipper-platform-cloudfoundry</module>
         <module>spring-cloud-skipper-platform-kubernetes</module>
+        <module>spring-cloud-skipper-autoconfigure</module>
+        <module>spring-cloud-skipper-dependencies</module>
+        <module>spring-cloud-starter-skipper-server</module>
     </modules>
     <dependencyManagement>
         <dependencies>
@@ -97,6 +100,11 @@
                 <version>${spring-cloud.version}</version>
                 <type>pom</type>
                 <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-skipper-autoconfigure</artifactId>
+                <version>1.0.2.BUILD-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>org.springframework.cloud</groupId>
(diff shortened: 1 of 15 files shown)


markpollack:
Should we rename `spring-cloud-skipper-autoconfigure` to `spring-cloud-skipper-server-autconfigure`?

pom.xml
```diff
@@ -98,6 +101,11 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-skipper-autoconfigure</artifactId>
+                <version>1.0.2.BUILD-SNAPSHOT</version>
```
markpollack: WDYT about replacing the literal version string with `${project.version}` ?

markpollack:
rebased and merged as 280a8809b48b1cb3059cf9e67d244d11e86aff55

pom.xml
jvalkeal: I don't remember why I went with actual version as at some point these were using `${project.version}`. Don't remember anymore if I had some issues with that. I had some many issues that I shuffled things around so much. We can try later and see if it works(it prob should be ok).

PR #585, opened 2018-04-12
Fix various issues around state handling

### LINKED ISSUE #574: Spring-Cloud-Skipper DefaultStateMachineService is not saved/persisted
Introduction
-

I have an issue of skipper is stopped. 
I have had packages that have been uploaded to the skipper service.

Package search displays the packages normally. 
[![enter image description here][1]][1]

  [1]: https://i.stack.imgur.com/vfh5V.png

However, I am then trying to `install` the specific packages using `org.springframework.cloud.skipper.client.DefaultSkipperClient` there is an error.

    SNAPSHOT:skipper-install failed: 
        org.springframework.web.client.HttpServerErrorException: 500 null -> [Help 1]

Error on the server:
-
**The error on the actual server is due `this.machines` being an empty map.**

    DefaultStateMachineService.acquireStateMachine(String machineId, boolean start).

However, this doesn't make sense if this map is not persisted and disappears after ever restart.

### LINKED ISSUE #575: Trying to delete a release but state is unknown
Introduction
-
I'm trying to delete a release and am getting a 500 error.
The application seems to be in State = [unknown].

[![enter image description here][1]][1]

  [1]: https://i.stack.imgur.com/xQYhS.png

The remote debugger to skipper-core instance shows that an exception is being thrown when the method `stateMachineService.acquireStateMachine(machineId);` .

Additionally, it seems that the State was unknown for the specific instances.

Question
-
Is there any way to purge/delete these release meta-data entries? Is there a way to recover from this?

jvalkeal:
- Effectively remove most of the release related object
  from machine context persistance layer to overcome
  kryo issues.
- Remove hard requirement passing Release pojo around
  and try to recreate it and report from request pojos
  used with install and upgrade.
- Rewrite most of the machine action classes to be better
  with serialised context when machine is restored.
- Add equals/hash to ConfigValues as tests expect equal on
  a object level, not just checking if it's a same instance.
- Some things are left to other PR's such as, 1. have proper
  release status when upgrading, 2. auto start machines for
  releases which are supposed to be in a state other than
  INITIAL, 3 need to add support for installing a release when
  upgrade fully failed where v1 ended into DELETED and v2 into FAILED.
- Fixes #574
- Fixes #575

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
@@ -135,7 +135,7 @@ public class AppDeployerReleaseManager implements ReleaseManager {
 	}
 
 	@Override
-	public ReleaseAnalysisReport createReport(Release existingRelease, Release replacingRelease) {
+	public ReleaseAnalysisReport createReport(Release existingRelease, Release replacingRelease, boolean initial) {
 		ReleaseAnalysisReport releaseAnalysisReport = this.releaseAnalyzer.analyze(existingRelease, replacingRelease);
 		if (releaseAnalysisReport.getReleaseDifference().areEqual()) {
 			throw new SkipperException(
@@ -155,7 +155,9 @@ public class AppDeployerReleaseManager implements ReleaseManager {
 		Manifest manifest = new Manifest();
 		manifest.setData(manifestData);
 		replacingRelease.setManifest(manifest);
-		this.releaseRepository.save(replacingRelease);
+		if (initial) {
+			this.releaseRepository.save(replacingRelease);
+		}
 		return releaseAnalysisReport;
 	}
 
@@ -275,7 +277,17 @@ public class AppDeployerReleaseManager implements ReleaseManager {
 		List<String> deploymentIds = appDeployerData.getDeploymentIds();
 		if (!deploymentIds.isEmpty()) {
 			for (String deploymentId : deploymentIds) {
-				appDeployer.undeploy(deploymentId);
+
+				// don't error trying trying to undeploy something
+				// which is not deployed
+				AppStatus appStatus = appDeployer.status(deploymentId);
+				if (appStatus.getState().equals(DeploymentState.deployed)) {
+					appDeployer.undeploy(deploymentId);
+				}
+				else {
+					logger.warn("For Release name {}, did not undeploy existing app {} as its status is not "
+									+ "'deployed'.", release.getName(), deploymentId);
+				}
 			}
 			Status deletedStatus = new Status();
 			deletedStatus.setStatusCode(StatusCode.DELETED);
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HandleHealthCheckStep.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HandleHealthCheckStep.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2017-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -105,9 +105,11 @@ public class HandleHealthCheckStep {
 			this.releaseRepository.save(replacingRelease);
 		}
 		catch (DataAccessException e) {
+			logger.debug("Error1 deleteReplacingRelease {}", e);
 			throw e;
 		}
 		catch (Exception e) {
+			logger.debug("Error2 deleteReplacingRelease {}", e);
 			// Update Status in DB
 			Status status = new Status();
 			status.setStatusCode(StatusCode.FAILED);
@@ -131,9 +133,11 @@ public class HandleHealthCheckStep {
 			this.deleteStep.delete(existingRelease, existingAppDeployerData, applicationNamesToUpgrade);
 		}
 		catch (DataAccessException e) {
+			logger.debug("Error1 deleteExistingRelease {}", e);
 			throw e;
 		}
 		catch (Exception e) {
+			logger.debug("Error2 deleteExistingRelease {}", e);
 			// Update Status in DB
 			Status status = new Status();
 			status.setStatusCode(StatusCode.FAILED);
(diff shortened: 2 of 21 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
```diff
@@ -275,7 +277,17 @@ public Release delete(Release release) {
 		List<String> deploymentIds = appDeployerData.getDeploymentIds();
 		if (!deploymentIds.isEmpty()) {
 			for (String deploymentId : deploymentIds) {
-				appDeployer.undeploy(deploymentId);
+
+				// don't error trying trying to undeploy something
+				// which is not deployed
+				AppStatus appStatus = appDeployer.status(deploymentId);
+				if (appStatus.getState().equals(DeploymentState.deployed)) {
```
ilayaperumalg: In case of multiple instances requested for the app, the appStatus could be `partial` and we expect to `undeploy` the ones that are deployed in that case as well.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HandleHealthCheckStep.java
```diff
@@ -105,9 +105,11 @@ private void deleteReplacingRelease(Release replacingRelease) {
 			this.releaseRepository.save(replacingRelease);
 		}
 		catch (DataAccessException e) {
+			logger.debug("Error1 deleteReplacingRelease {}", e);
```
ilayaperumalg: Did you add this for some debugging purpose or do we want this to be like this here?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/statemachine/RollbackStartAction.java
```diff
@@ -54,9 +56,17 @@ public RollbackStartAction(ReleaseRepository releaseRepository) {
 
 	@Override
 	protected void executeInternal(StateContext<SkipperStates, SkipperEvents> context) {
+		log.debug("Starting to execute action");
```
ilayaperumalg: Could this be, `Starting to execute rollback action`

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HandleHealthCheckStep.java
jvalkeal: Yes, there's Error1 and Error2 in two other places for debugging purposes. We might switch those to trace level?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/statemachine/RollbackStartAction.java
jvalkeal: Doesn't matter, logging will tell class where debug happens.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
jvalkeal: This was copied from `DeleteStep` as is to check same case as otherwise call to undeploy would fail. Would you complain same thing in `DeleteStep` as well?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HandleHealthCheckStep.java
```diff
@@ -131,9 +133,11 @@ private void deleteExistingRelease(Release existingRelease, List<String> applica
 			this.deleteStep.delete(existingRelease, existingAppDeployerData, applicationNamesToUpgrade);
 		}
 		catch (DataAccessException e) {
+			logger.debug("Error1 deleteExistingRelease {}", e);
```
markpollack: Maybe `Database Error` instead of the generic `Error1`  Why suppress this completely, e.g. not at a warning level.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HandleHealthCheckStep.java
```diff
@@ -131,9 +133,11 @@ private void deleteExistingRelease(Release existingRelease, List<String> applica
...
 			throw e;
 		}
 		catch (Exception e) {
+			logger.debug("Error2 deleteExistingRelease {}", e);
```
markpollack: Something better than `Error2` ... also same comment wrt to completely suppressing

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/strategies/HandleHealthCheckStep.java
jvalkeal: I think these exceptions only have meaning on a debug level to know if it was first or second catch. That's essentially why I added these on debug level.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/statemachine/RollbackStartAction.java
```diff
@@ -84,20 +94,39 @@ protected void executeInternal(StateContext<SkipperStates, SkipperEvents> contex
...
+		else {
+			UpgradeRequest upgradeRequest = new UpgradeRequest();
+			UpgradeProperties upgradeProperties = new UpgradeProperties();
+			upgradeProperties.setReleaseName(releaseName);
+			upgradeProperties.setConfigValues(releaseToRollback.getConfigValues());
+			upgradeRequest.setUpgradeProperties(upgradeProperties);
+			upgradeRequest.setPackageIdentifier(packageIdentifier);
+			context.getExtendedState().getVariables().put(SkipperEventHeaders.UPGRADE_REQUEST, upgradeRequest);
+		}
+		// TODO: originally we had
```
markpollack: can remove this on merge.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/AppDeployerReleaseManager.java
```diff
@@ -155,7 +155,9 @@ public ReleaseAnalysisReport createReport(Release existingRelease, Release repla
 		Manifest manifest = new Manifest();
 		manifest.setData(manifestData);
 		replacingRelease.setManifest(manifest);
-		this.releaseRepository.save(replacingRelease);
+		if (initial) {
+			this.releaseRepository.save(replacingRelease);
```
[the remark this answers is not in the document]
jvalkeal: Yeah, that would allow us to remove that boolean flag. It's cleaner if report doesn't store anything.

Issue #561, opened 2018-03-14
Handle required autoconfig excludes


jvalkeal:
Default fatjar app uses
```
@SpringBootApplication(exclude = {
		CloudFoundryDeployerAutoConfiguration.class,
		KubernetesAutoConfiguration.class,
		LocalDeployerAutoConfiguration.class,
		ManagementWebSecurityAutoConfiguration.class,
		SecurityAutoConfiguration.class,
		SessionAutoConfiguration.class
	})
@EnableSkipperServer
public class SkipperServerApplication {
```
It is a bit nasty to except user to create a custom skipper app and handle same exclude logic.

jvalkeal:
What comes for autoconfigs for local/cf/k8s, we could separate autoconfigs into their own jars and then in skipper would need to include only deployer core's.

Security and session is more problematic as those are force fed to us and we had same dance in dataflow with `SessionAutoConfiguration` where we decided to use property to disable session autoconfig.

PR #591, opened 2018-04-23
Add instructions for custom server build

### LINKED ISSUE #552: Document how to run a customized version of spring-cloud-skipper-server
(brought this up in gitter, @sabbyanandan recommended I open an issue)

The documentation mentions adding a custom `AuthoritiesExtractor` in the security section, but all the documentation about deploying skipper only covers downloading a release jar and running it directly or using a docker image.

It is unclear how to set up a skipper server with modified beans/other customizations.

In similar projects it might be something like
* create a new spring boot app
* add a skipper-server dependency
* add a `@EnabledSkipperServer` annotation 
but currently,  it is not clear how to do that.

jvalkeal:
- Example of a main class, maven and gradle build files.
- Fixes #552

### DIFF
--- a/spring-cloud-skipper-docs/src/main/asciidoc/appendix-building.adoc
+++ b/spring-cloud-skipper-docs/src/main/asciidoc/appendix-building.adoc
@@ -14,7 +14,7 @@ $ ./mvnw clean install
 To create the executables and avoid running the tests and generating JavaDocs, use the following command:
 
 ----
-$./mvnw clean package -DskipTests -Dmaven.javadoc.skip=true
+$ ./mvnw clean package -DskipTests -Dmaven.javadoc.skip=true
 ----
 
 NOTE: You can also install Maven (>=3.3.3) yourself and run the `mvn` command in place of `./mvnw` in the examples.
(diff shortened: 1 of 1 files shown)


spring-cloud-skipper-docs/src/main/asciidoc/appendix-building.adoc
```diff
@@ -28,21 +28,208 @@ We try to cover this in the `.mvn` configuration, so, if you find you have to in
...
+import org.springframework.boot.actuate.autoconfigure.ManagementWebSecurityAutoConfiguration;
+import org.springframework.boot.autoconfigure.security.SecurityAutoConfiguration;
+import org.springframework.boot.autoconfigure.session.SessionAutoConfiguration;
+import org.springframework.cloud.deployer.spi.cloudfoundry.CloudFoundryDeployerAutoConfiguration;
+import org.springframework.cloud.deployer.spi.kubernetes.KubernetesAutoConfiguration;
+import org.springframework.cloud.deployer.spi.local.LocalDeployerAutoConfiguration;
+import org.springframework.cloud.skipper.server.EnableSkipperServer;
+
+@SpringBootApplication(exclude = {
+                CloudFoundryDeployerAutoConfiguration.class,
```
ilayaperumalg: Can we add a note on why we exclude these auto configuration classes?

jvalkeal: I wonder what should we write other than all these auto-configs simply break skipper if not excluded?

jvalkeal: Actually we have #561 which is a bit broader issue around this fact that these needs to be excluded.

PR #599, opened 2018-04-25
Add default deployer for single deployer platform

### LINKED ISSUE #543: Implement auto-election of "default" platform
As a user, I'm trying to publish a package to CF; however, I get the following error in return.

```
skipper:>platform list
╔════╤════════════╤══════════════════════════════════════════════════════════════════════════╗
║Name│    Type    │                               Description                                ║
╠════╪════════════╪══════════════════════════════════════════════════════════════════════════╣
║dev │cloudfoundry│org = [scdf-ci], space = [space-sabby], url = [https://api.run.pivotal.io]║
╚════╧════════════╧══════════════════════════════════════════════════════════════════════════╝

skipper:>package install --package-name helloworld  --release-name somepackage
No deployer named 'default'
Details of the error have been omitted. You can use the stacktrace command to print the full stacktrace.
```

Note, I've specifically disabled the local-deployer account (via `SPRING_CLOUD_SKIPPER_SERVER_ENABLELOCALPLATFORM: false`). I wanted `dev` to be the default platform, but it is not picked up automatically.

ilayaperumalg:
- If there is a single platform that exists with a single deployer,
then create a `default` deployer using the existing deployer.
 - This will allow user to use the default deployer without having to specify the `platformName`

Resolves #543

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/DeployerInitializationService.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/DeployerInitializationService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2017-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/DeployerInitializationService.java
```diff
@@ -61,4 +81,12 @@ public void initialize(ApplicationReadyEvent event) {
 			});
 		});
 	}
+
+	private boolean singleDeployerExists() {
+		int deployersCount = 0;
+		for (Platform platform: this.platforms) {
+			deployersCount = deployersCount + platform.getDeployers().size();
+		}
+		return (deployersCount > 1) ? false : true;
```
jvalkeal: Just to make sure, if `deployersCount` is still zero at this point, do logic around `singleDeployerExists()` still work?

jvalkeal:
I think this looks ok. I'll do a bit of poking around tomorrow morning and merge it.

jvalkeal:
Merged per 7e497208986b5ca6d3997b16c873bbb8f43cd1d7

PR #608, opened 2018-05-04
Enhance timeout handling with rollback and upgrade

### LINKED ISSUE #587: Add parameter for upgrade timeout.
Currently timeout is a global parameter in as server. It would be nice if client operation could pass this parameter along with upgrade request.

`package install`, `release delete`, `release rollback` and `release upgrade` could have option `--timeout` which would implicate that if there is an operation of some sort for a particular ongoing flow or deployer, timeout could be used.

### LINKED ISSUE #598: Implement Boot 2.x style duration expressions
We want to have same duration expression as supported in boot so that when we finally upgrade to boot 2.x, expressions or time definitions we used would be supported as is.

For example, with boot 2.x you can use strings like `500ms`, `PT10S`, `PT6M30S`, etc.

child of #597

### LINKED ISSUE #601: Change rollback api to use domain object
With install we use `InstallRequest` and with upgrade we use `UpgradeRequest` which allows to extend stuff passed in, i.e. for timeout field.

Rollback simply uses uri parameters to pass in `releaseName` and `releaseVersion` which doesn't work in a long run when we need to pass in more options. Should replace api with `RollbackRequest`.

child of #597

jvalkeal:
- This commit focus to add better functionality to
  handle timeouts during upgrade and rollback.
- We start by adding `timeout-expression` option to both
  `release upgrade` and `release rollback`. This expression
  is backed by a feature copy of `DurationUtils` and `DurationStyle`
  from boot2, which gives nice upgrade path if that time comes.
  Behind a scenes these expressions are translated to millis.
- New duration expression takes all sort of strings like, `60s`,
  `PT1H`, etc.
- As we need to pass timout with rollback, introduce `RollbackRequest`
  and deprecate previous endpoint which only took `releaseName`
  and `version`.
- `ReleaseController` now uses `/rollback` to get `RollbackRequest`,
   while `/rollback/{name}/{version}` is the old deprecated endpoint.
- Add some new tests to statemachine. Core timeout functionality
  were actually already in place, just not exposed to user.
- Modify `UpgradeStrategy.cancel` to pass timeout value so that
  we can report it back to status as timeout can come from user request.
- Fix bug where timeout from `HealthCheckProperties` were not
  actually used as a global setting. We previously had hard coded
  value in a statemachine. Now order is, shell/rest timeout value ->
  HealthCheckProperties -> default value in statemachine.
- Add new rest docs for new api and add note for deprecation.
- Some status fields don't exactly represent correct operation now,
  something we need to fix in other commits. We used to manually change
  status from rollback to indicate rollback even if we're dispatching to
  install or upgrade. This is a bit broader topic we need to address as
  we can't just manually change things outside of the flow.
- Fixes #587
- Fixes #598
- Fixes #601

### DIFF
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
@@ -35,6 +35,7 @@ import org.springframework.cloud.skipper.domain.Manifest;
 import org.springframework.cloud.skipper.domain.PackageMetadata;
 import org.springframework.cloud.skipper.domain.Release;
 import org.springframework.cloud.skipper.domain.Repository;
+import org.springframework.cloud.skipper.domain.RollbackRequest;
 import org.springframework.cloud.skipper.domain.Template;
 import org.springframework.cloud.skipper.domain.UpgradeRequest;
 import org.springframework.cloud.skipper.domain.UploadRequest;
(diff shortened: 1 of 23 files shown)


ilayaperumalg:
LGTM

ilayaperumalg:
Merging.

PR #643, opened 2018-05-23
Handle package config value YAML loading

ilayaperumalg:
- Remove double quotes in generic template for the property values
 - Perform Yaml dump on merged config values with the YAML dumping option set to DumperOptions.ScalarStyle.DOUBLE_QUOTED for the default scalar style
 - Load this yaml data when setting the release manifest

Resolves #624

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/ReleaseService.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/ReleaseService.java
@@ -15,12 +15,15 @@
  */
 package org.springframework.cloud.skipper.server.service;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.DumperOptions;
+import org.yaml.snakeyaml.Yaml;
 
 import org.springframework.cloud.skipper.PackageDeleteException;
 import org.springframework.cloud.skipper.ReleaseNotFoundException;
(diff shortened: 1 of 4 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/ReleaseService.java
```diff
@@ -185,9 +188,18 @@ public Release install(Release release) {
 		Map<String, Object> mergedMap = ConfigValueUtils.mergeConfigValues(release.getPkg(), release.getConfigValues());
 		// Render yaml resources
 		String manifestData = ManifestUtils.createManifest(release.getPkg(), mergedMap);
+		DumperOptions dumperOptions = new DumperOptions();
+		dumperOptions.setDefaultScalarStyle(DumperOptions.ScalarStyle.DOUBLE_QUOTED);
+		dumperOptions.setPrettyFlow(true);
```
tzolov: Don't think that pretty print is necessary here.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/ReleaseService.java
```diff
@@ -185,9 +188,18 @@ public Release install(Release release) {
 		Map<String, Object> mergedMap = ConfigValueUtils.mergeConfigValues(release.getPkg(), release.getConfigValues());
 		// Render yaml resources
 		String manifestData = ManifestUtils.createManifest(release.getPkg(), mergedMap);
+		DumperOptions dumperOptions = new DumperOptions();
```
tzolov: Wouldn't the ManifestUtils.createManifest be a better place to keeps the  YAML reload/dump gymnastic?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/ReleaseService.java
```diff
@@ -185,9 +188,18 @@ public Release install(Release release) {
...
+		dumperOptions.setDefaultScalarStyle(DumperOptions.ScalarStyle.DOUBLE_QUOTED);
+		dumperOptions.setPrettyFlow(true);
+		Yaml yaml = new Yaml(dumperOptions);
+		Iterable<Object> loadedYaml = yaml.loadAll(manifestData);
 		logger.debug("Manifest = " + ArgumentSanitizer.sanitizeYml(manifestData));
```
tzolov: This is obsolete. The dumped manifest should be logged instead.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/ReleaseService.java
markpollack: yea, saves a bit of space as it gets stored in the DB, but maybe helps with debugging?

ilayaperumalg:
This can be closed as #644 is based on top of this.

PR #659, opened 2018-05-31
Fix typo in persist skip function

### LINKED ISSUE #658: Wrong extended state persist filtering
Looks like `SkipUnwantedVariablesFunction` contains a typo and thus `ERROR` leaks into persisted context and thus may cause issues if error itself cannot be de-serialised with kryo.

jvalkeal:
- Expose SkipUnwantedVariablesFunction to package level
  to test it.
- Fix typo in apply function which caused wrong check
  for SkipperVariables.ERROR.
- Fixes #658


markpollack:
merged in master as b01a5a4d83b4acaaceef2710146e1a3aca501e43 and 1.0.x as f9c7c2058f7735fc6fec0b9dd4632177e224cdb2

PR #660, opened 2018-05-31
Add quotes back to generic-template.yml

markpollack:
Should also be added onto master branch


markpollack:
Merged on master as d3190311b8e585f31343b55d66ebe906cc9343cd and on 1.0.x as 474f92cf57b980eeaa018fbf465d0105d16936ba

PR #666, opened 2018-06-07
Update test app to use random ports

### LINKED ISSUE #636: CI tests flaky
We're getting random port bind errors and reason seem to be we not explicitly not setting `server.port` to zero. If this setting doesn't exist, local deployer then trying to use unreliable socket utils to find free port and then manually sets that port. Either we need to manually modify all existing apps used in skipper or from tests actually set `server.port=0`.

jvalkeal:
- Basically for ticktock apps which are most likely
  to fail with port conflights in local, outside of
  docker, start using `server.port=0`.
- Fixes #636


ilayaperumalg:
LGTM, merging.

PR #668, opened 2018-06-14
Refactor to support multiple release managers

### LINKED ISSUE #629: Support multiple release managers
Instead of using `DefaultReleaseManager` or `AppDeployerReleaseManager` or any other mean to dispatch manifest to correct manager, we should allow user to register individual release managers and then manager would have method

```
boolean supportKind(String kind);
```

child of #627

jvalkeal:
- spring-cloud-skipper-platform-kubernetes now depends
  on spring-cloud-skipper-server-core.
- New interfaces ReleaseManagerFactory and UpgradeStrategyFactory
  which can return particular instances per application kind, which
  then allows to move cf manifest classes out from server core
  to cf platform package.
- Rest of the changes are pretty much refactoring and moving
  stuff around.
- Fixes #629

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
@@ -17,6 +17,7 @@ package org.springframework.cloud.skipper.server.config;
 
 import java.io.File;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executor;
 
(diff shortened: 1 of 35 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerConfiguration.java
```diff
@@ -36,9 +37,10 @@
 import org.springframework.cloud.deployer.resource.maven.MavenResourceLoader;
 import org.springframework.cloud.deployer.resource.support.DelegatingResourceLoader;
 import org.springframework.cloud.deployer.resource.support.LRUCleaningResourceLoaderBeanPostProcessor;
-import org.springframework.cloud.skipper.deployer.cloudfoundry.CloudFoundryPlatformProperties;
-import org.springframework.cloud.skipper.deployer.cloudfoundry.PlatformCloudFoundryOperations;
-import org.springframework.cloud.skipper.domain.CFApplicationManifestReader;
+//import org.springframework.cloud.skipper.deployer.cloudfoundry.CFManifestApplicationDeployer;
```
ilayaperumalg: We can remove these.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 2683d6e

PR #670, opened 2018-06-19
Handle deletion of release that is being installed

### LINKED ISSUE #669: Deletion of release while it is being installed
If the release deletion request is received while the install request is being processed, the deletion never happens and subsequent release status would use the app deployer data of this release.

ilayaperumalg:
- If the deletion of release request is sent while the release is being installed, the deletion has to wait until the state changes.
This will make sure we update the deletion correctly.

Resolves #669


ilayaperumalg:
Closing this in favor of other PR.

PR #671, opened 2018-06-20
Release deletion always undeploys the apps

### LINKED ISSUE #669: Deletion of release while it is being installed

ilayaperumalg:
- When the release deletion is requested, the apps associated with the release are always deleted irrespective of their deployment status.
 - When the status is requested for the release in `DELETED` state, return the release as is (without any further computation of deployment)

Resolves #669

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/DefaultReleaseManager.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/DefaultReleaseManager.java
@@ -85,7 +85,7 @@ public class DefaultReleaseManager implements ReleaseManager {
 		this.appDeploymentRequestFactory = appDeploymentRequestFactory;
 		this.applicationManifestReader = applicationManifestReader;
 	}
-	
+
 	@Override
 	public Collection<String> getSupportedKinds() {
 		return Arrays.asList(SkipperManifestKind.SpringBootApp.name(),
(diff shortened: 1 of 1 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/DefaultReleaseManager.java
```diff
@@ -282,28 +286,22 @@ public Release status(Release release) {
...
-				AppStatus appStatus = appDeployer.status(deploymentId);
-				if (appStatus.getState().equals(DeploymentState.deployed)) {
+				try  {
 					appDeployer.undeploy(deploymentId);
 				}
-				else {
-					logger.warn("For Release name {}, did not undeploy existing app {} as its status is not "
-							+ "'deployed'.", release.getName(), deploymentId);
+				catch (Exception e) {
+					throw new SkipperException(e.getMessage());
```
markpollack: we should not be throwing an exception here, we need to iterate over all the IDs and log errors, otherwise we may end up in an inconsistent state with some apps not undeployed.

ilayaperumalg: oh yeah, I will update the PR.

markpollack:
Merged as 162ef7f763dd89a55a17bf60e952e17d934dd2af

PR #678, opened 2018-07-03
Add declarative configuration of DataSource in cloud profile [needs work]

### LINKED ISSUE #677: Add declarative configuration of DataSource in cloud profile
for setting connection pool size when on PCF

markpollack:
CF is still picking up a tomcat connection pool.  not sure why.

Resolves #677

### DIFF
--- a/spring-cloud-skipper-server-core/pom.xml
+++ b/spring-cloud-skipper-server-core/pom.xml
@@ -182,6 +182,22 @@
         </dependency>
         <dependency>
             <groupId>io.pivotal.spring.cloud</groupId>
+            <artifactId>spring-cloud-services-starter-config-client</artifactId>
+            <version>1.6.3.RELEASE</version>
+            <scope>runtime</scope>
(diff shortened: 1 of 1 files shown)


spring-cloud-skipper-server-core/pom.xml
```diff
@@ -14,6 +14,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>3.1.0</version>
```
jvalkeal: Can we put all new deps to dep management in parent pom? Same in `spring-cloud-skipper-server`

sabbyanandan:
See https://github.com/spring-cloud/spring-cloud-dataflow-server-cloudfoundry/pull/416 for instructions.

jvalkeal:
ok, version defs into parent pom and then merging.

PR #687, opened 2018-07-18
Test SkipperServer to allow cmdline args

### LINKED ISSUE #686: Test Skipper server needs to handle command line arguments
The current `LocalSkipperResource` from `spring-cloud-skipper-server` tests hard codes the `server.port=0` and makes it as the only possible cmd line argument to be passed to the Skipper server.

We need to expand it to support passing command line arguments.

ilayaperumalg:
- The `LocalSkipperResource` in `spring-cloud-skipper-server` tests now has the option to pass the command line arguments

Resolves #686


ilayaperumalg:
Merged into master as e1d566bc9060594936c2446970ed685564fe158e and 1.0.x as 1fc651c

Issue #691, opened 2018-07-23
CF Manifest property handling should be more clever


jvalkeal:
Using same `memory` format from yml breaks substitution as only `Integer` is supported
```
skipper:>package install --package-name jannepython-printer --package-version 1.0.0 --platform-name cf-dev --release-name my-jannepython-printer --properties "spec.manifest.memory=2G"
```
For this you need to use `Integer`:
```
skipper:>package install --package-name jannepython-printer --package-version 1.0.0 --platform-name cf-dev --release-name my-jannepython-printer --properties "spec.manifest.memory=2048"
```

PR #699, opened 2018-07-31
Move from manifest file to spec format

### LINKED ISSUE #695: Overhaul CF Manifest file handling
1. I think we need to be strict how additional files are processed. I.e in this case manifest deployer is simply trying to find first file ending with `manifest.yml`. Then values are updated from template.

2. This have few issues like install blows up if manifest file is not it place. Also as manifest file is same format as one from CF, only first application entry is used. All these feels rather awkward.

Maybe it'd be better to have a source of thruth in template like:

```
apiVersion: skipper.spring.io/v1
kind: CloudFoundryApplication
spec:
  manifest:
    name: jannelogcf
    memory: 1024
    disk_quota: 1024
    timeout: 180
    instances: 1
    path: log-sink-rabbit-1.1.0.RELEASE.jar
    services: rabbit

```

and then if we really want to insert something into `spec.manifest`, maybe we can extend mustache templates to read/export additional files from a package.

child of #627

jvalkeal:
- Remove use of cf manifest file in favour
  of using a spec itself.
- Fix all bundled samples
- Fixes #695
- Breaks #691 as use of YamlMapper should get
  overhauled as all other property key types
  and values should be handled with YamlMapper
  customization.


spring-cloud-skipper-platform-cloudfoundry/src/main/java/org/springframework/cloud/skipper/deployer/cloudfoundry/CloudFoundryManifestApplicationDeployer.java
```diff
@@ -79,52 +79,49 @@ public CloudFoundryManifestApplicationDeployer(CloudFoundryApplicationManifestRe
...
+					.healthCheckType(manifest.getHealthCheckType() != null
+							? ApplicationHealthCheck.from(manifest.getHealthCheckType().name())
+							: null)
+					.hosts(manifest.getHosts() != null && !manifest.getHosts().isEmpty() ? manifest.getHosts() : null)
+					.instances(manifest.getInstances())
+					.memory(CloudFoundryApplicationManifestUtils.memoryInteger(manifest.getMemory()))
+					.noHostname(manifest.getNoHostname())
+					.noRoute(manifest.getNoRoute())
+					.randomRoute(manifest.getRandomRoute())
+					.services(manifest.getServices())
```
ilayaperumalg: We need to have some handlers for these properties that convert properties into the appropriate type.

For instance,

When we do 
```
skipper:>release upgrade --release-name a1 --package-name logcf --package-version 1.0.1 --properties "spec.manifest.memory=2048,spec.manifest.services=mysql,rabbit"
```

The services values are expected to be updated. But, we get this:

```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: Can not deserialize instance of java.util.ArrayList out of VALUE_STRING token
 at [Source: java.io.StringReader@7026f39b; line: 10, column: 17] (through reference chain: org.springframework.cloud.skipper.domain.CloudFoundryApplicationSkipperManifest["spec"]->org.springframework.cloud.skipper.domain.CloudFoundryApplicationSpec["manifest"]->org.springframework.cloud.skipper.domain.CloudFoundryApplicationSpec$Manifest["services"])
	at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:270) ~[jackson-databind-2.8.11.2.jar!/:2.8.11.2]
	at com.fasterxml.jackson.databind.DeserializationContext.reportMappingException(DeserializationContext.java:1247) ~[jackson-databind-2.8.11.2.jar!/:2.8.11.2]
```

jvalkeal: Don't know if `spec.manifest.services=[mysql,rabbit]` would actually work. You need to remember that we're replacing yml and `mysql,rabbit` cannot be mapped to list. Essentially you're creating broken yml.

jvalkeal: Yeah, this did work:
```
package install --package-name logcf --release-name mylogcf --platform-name cf-dev --package-version 1.0.0 --properties "spec.manifest.memory=2048,spec.manifest.services=[mysql,rabbit]"
```

ilayaperumalg: yeah, The `upgrade` is still broken even when we specify with `[]`

```
java.lang.NullPointerException: null
	at org.springframework.cloud.skipper.deployer.cloudfoundry.CloudFoundryApplicationManifestUtils.getCFManifestMap(CloudFoundryApplicationManifestUtils.java:79) ~[spring-cloud-skipper-platform-cloudfoundry-1.1.0.BUILD-SNAPSHOT.jar!/:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.skipper.deployer.cloudfoundry.CloudFoundryReleaseAnalyzer.analyze(CloudFoundryReleaseAnalyzer.java:67) ~[spring-cloud-skipper-platform-cloudfoundry-1.1.0.BUILD-SNAPSHOT.jar!/:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.skipper.deployer.cloudfoundry.CloudFoundryReleaseManager.createReport(CloudFoundryReleaseManager.java:146) ~[spring-cloud-skipper-platform-cloudfoundry-1.1.0.BUILD-SNAPSHOT.jar!/:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.skipper.server.service.ReleaseReportService.createReport(ReleaseReportService.java:118) ~[spring-cloud-skipper-server-core-1.1.0.BUILD-SNAPSHOT.jar!/:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.skipper.server.service.ReleaseReportService$$FastClassBySpringCGLIB$$6bf7cb0e.invoke(<generated>) ~[spring-cloud-skipper-server-core-1.1.0.BUILD-SNAPSHOT.jar!/:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204) ~[spring-core-4.3.18.RELEASE.jar!/:4.3.18.RELEASE]
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:736) ~[spring-aop-4.3.18.RELEASE.jar!/:4.3.18.RELEASE]
```

ilayaperumalg:
Merging this PR as the issues identified from this PR are now tracked as #701 and #702.

ilayaperumalg:
Rebased and merged as a556372

PR #705, opened 2018-08-06
Fix manifest map handling

### LINKED ISSUE #702: CF Release Analyzer breaks when the comma separated list is provided
When the CF manifest properties of type `List` is provided as a comma-separated list within [], the CF release analyzer breaks.

For instance, if the manifest's `services` property has [rabbit,mysql] then this results in error mentioned [here](https://github.com/spring-cloud/spring-cloud-skipper/pull/699#discussion_r207094748)

jvalkeal:
- Now trying to build map manually which
  should not fail if special characters like
  commas are used in values.
- Fixes #702


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as acd153e

PR #707, opened 2018-08-06
Handle version for non maven/docker resources

### LINKED ISSUE #706: Handling version for non maven/docker resources
When loading non-maven/docker resources, the version option isn't handled as part of the resource.

ilayaperumalg:
- When loading the resources that are not maven or docker, then the version is expected to be added/substituted via `spec.version` in the template config
 - Update python packages
 - Update test

Resolves #706


spring-cloud-skipper-server-core/src/test/resources/repositories/sources/test/python/python-printer-1.0.1/templates/vanakkam.yml
```diff
@@ -1,7 +1,7 @@
 apiVersion: skipper.spring.io/v1
 kind: CloudFoundryApplication
 spec:
-  resource: "{{{spec.resource}}}"
+  resource: https://github.com/ilayaperumalg/sandbox/raw/master/python/1.0.1/hello.py-{{spec.version}}.zip
```
jvalkeal: I suppose to use version in a real life would not work because you have hardcoded base dirs, `1.0.0` and `1.0.1`.

jvalkeal:
Btw, I think you have a stale branch as things gets a bit crazy with merge conflicts.

jvalkeal:
Anyway, changes itself look ok.

PR #715, opened 2018-09-05
Add force upgrade option

### LINKED ISSUE #714: Add a "--force" option for package upgrades
As a user, I'd like to have a `--force` option for `skipper upgrade` command, so I can force upgrade the package even if there's no metadata change associated with it.

Notes:
This is a continuation from spring-cloud/spring-cloud-skipper#700. The idea is to allow upgrades for scenarios that involve config-server, where Skipper wouldn't be able to have direct access to query and react to external updates. The `--force` option would provide the ability to upgrade the package; likewise, the App would get rolling-upgraded, to bootstrap with the latest config-server changes.

ilayaperumalg:
- Add force upgrade option during release upgrade
   - When forced, the upgrade proceeds even if there are no differences
   - Allow users to set `appNames` to explicitly specify during upgrade when `force` option is used
 - Add command options
 - Update tests

Resolves #714

### DIFF
--- a/spring-cloud-skipper-platform-cloudfoundry/src/main/java/org/springframework/cloud/skipper/deployer/cloudfoundry/CloudFoundryReleaseAnalyzer.java
+++ b/spring-cloud-skipper-platform-cloudfoundry/src/main/java/org/springframework/cloud/skipper/deployer/cloudfoundry/CloudFoundryReleaseAnalyzer.java
@@ -16,6 +16,7 @@
 package org.springframework.cloud.skipper.deployer.cloudfoundry;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -56,9 +57,10 @@ public class CloudFoundryReleaseAnalyzer {
 	 * @param existingRelease the release that is currently deployed
 	 * @param replacingRelease the proposed release to be deployed that will replace the
 	 * existing release.
+	 * @param isForceUpdate flag to indicate if the update is forced
 	 * @return an analysis report describing the changes to make, if any.
 	 */
-	public ReleaseAnalysisReport analyze(Release existingRelease, Release replacingRelease) {
+	public ReleaseAnalysisReport analyze(Release existingRelease, Release replacingRelease, boolean isForceUpdate) {
 		List<ApplicationManifestDifference> applicationManifestDifferences = new ArrayList<>();
 		ApplicationManifest existingApplicationManifest = this.cfManifestApplicationDeployer.getCFApplicationManifest(existingRelease);
 		ApplicationManifest replacingApplicationManifest = this.cfManifestApplicationDeployer.getCFApplicationManifest(replacingRelease);
@@ -71,12 +73,14 @@ public class CloudFoundryReleaseAnalyzer {
 					emptyPropertiesDiff, emptyPropertiesDiff, emptyPropertiesDiff, propertiesDiff, emptyPropertiesDiff);
 			applicationManifestDifferences.add(applicationManifestDifference);
 		}
-		return createReleaseAnalysisReport(existingRelease, replacingRelease, applicationManifestDifferences);
+		return createReleaseAnalysisReport(existingRelease, replacingRelease, applicationManifestDifferences,
+				Arrays.asList(existingApplicationManifest.getName()), isForceUpdate);
 	}
 
 	private ReleaseAnalysisReport createReleaseAnalysisReport(Release existingRelease,
 			Release replacingRelease,
-			List<ApplicationManifestDifference> applicationManifestDifferences) {
+			List<ApplicationManifestDifference> applicationManifestDifferences, List<String> allApplicationNames,
+			boolean isForceUpdate) {
 		List<String> appsToUpgrade = new ArrayList<>();
 		ReleaseDifference releaseDifference = new ReleaseDifference();
 		releaseDifference.setDifferences(applicationManifestDifferences);
@@ -86,6 +90,7 @@ public class CloudFoundryReleaseAnalyzer {
 					StringUtils.collectionToCommaDelimitedString(releaseDifference.getChangedApplicationNames()) + "]");
 			appsToUpgrade.addAll(releaseDifference.getChangedApplicationNames());
 		}
-		return new ReleaseAnalysisReport(appsToUpgrade, releaseDifference, existingRelease, replacingRelease);
+		return new ReleaseAnalysisReport(appsToUpgrade, releaseDifference, existingRelease, replacingRelease,
+				allApplicationNames, isForceUpdate);
 	}
 }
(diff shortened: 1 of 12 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
```diff
@@ -61,9 +62,12 @@ public ReleaseAnalyzer(SpringCloudDeployerApplicationManifestReader applicationM
 	 * @param existingRelease the release that is currently deployed
 	 * @param replacingRelease the proposed release to be deployed that will replace the
 	 * existing release.
+	 * @param isForceUpdate flag to indicate if the update is forced
+	 * @param appNamesToUpdate the application names to force update
 	 * @return an analysis report describing the changes to make, if any.
 	 */
-	public ReleaseAnalysisReport analyze(Release existingRelease, Release replacingRelease) {
+	public ReleaseAnalysisReport analyze(Release existingRelease, Release replacingRelease, boolean isForceUpdate,
+			String[] appNamesToUpdate) {
```
markpollack: List instead of array?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
```diff
@@ -168,7 +190,12 @@ private ReleaseAnalysisReport createReleaseAnalysisReport(Release existingReleas
 					StringUtils.collectionToCommaDelimitedString(releaseDifference.getChangedApplicationNames()) + "]");
 			appsToUpgrade.addAll(releaseDifference.getChangedApplicationNames());
 		}
-		return new ReleaseAnalysisReport(appsToUpgrade, releaseDifference, existingRelease, replacingRelease);
+		// Set explicit app names from the client only if the update is enforced.
+		else if (isForceUpdate && appNamesToUpdate != null) {
```
markpollack: If a stream update also passes in properties, then the force option is ignored.  I guess we should take into account that some apps in the stream maybe updated due the stream update changing properties in addition to some apps being listed in the force option.  Otherwise, we should make the command for stream update (from SCDF perspective) have mutually exclusive command arguments (properties and force).

markpollack: I played around with it, and it gets sort of complicated allowing both a `--propertiesFile` to be specified in addition to a `--force` option as well.  For example, in `ReleaseAnalysisReport` there is the code 

```
	public List<String> getApplicationNamesToUpgrade() {
		if (this.isForceUpdate) {
			return this.applicationNamesToUpgrade.size() > 0 ? this.applicationNamesToUpgrade : this.allApplicationNames;
		}
		else{
			return this.applicationNamesToUpgrade;
		}
	}
```

So given `http | tx | log`,  if we pass `--propertiesFile` for the `http` app and also `--force` then only `http` will be updated.  I think we need to introduce a new command `stream update-force` and corresponding new Controller that only accepts those options.  This makes the contract clear.

markpollack: I think I found a way to make the single command have a clear contract....

spring-cloud-skipper-platform-cloudfoundry/src/main/java/org/springframework/cloud/skipper/deployer/cloudfoundry/CloudFoundryReleaseAnalyzer.java
```diff
@@ -71,12 +73,14 @@ public ReleaseAnalysisReport analyze(Release existingRelease, Release replacingR
 					emptyPropertiesDiff, emptyPropertiesDiff, emptyPropertiesDiff, propertiesDiff, emptyPropertiesDiff);
 			applicationManifestDifferences.add(applicationManifestDifference);
 		}
-		return createReleaseAnalysisReport(existingRelease, replacingRelease, applicationManifestDifferences);
+		return createReleaseAnalysisReport(existingRelease, replacingRelease, applicationManifestDifferences,
+				Arrays.asList(existingApplicationManifest.getName()), isForceUpdate);
```
markpollack: I am not sure this works, isn't `Arrays.asList(existingApplicationManifest.getName())` going to return just a single name instead of 'all application names'?

spring-cloud-skipper-platform-cloudfoundry/src/main/java/org/springframework/cloud/skipper/deployer/cloudfoundry/CloudFoundryReleaseAnalyzer.java
```diff
@@ -56,9 +57,10 @@ public CloudFoundryReleaseAnalyzer(CloudFoundryManifestApplicationDeployer cfMan
 	 * @param existingRelease the release that is currently deployed
 	 * @param replacingRelease the proposed release to be deployed that will replace the
 	 * existing release.
+	 * @param isForceUpdate flag to indicate if the update is forced
 	 * @return an analysis report describing the changes to make, if any.
 	 */
-	public ReleaseAnalysisReport analyze(Release existingRelease, Release replacingRelease) {
+	public ReleaseAnalysisReport analyze(Release existingRelease, Release replacingRelease, boolean isForceUpdate) {
```
markpollack: why isn't the list of appNamesToUpdate passed in here?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
ilayaperumalg: Sure!

spring-cloud-skipper-platform-cloudfoundry/src/main/java/org/springframework/cloud/skipper/deployer/cloudfoundry/CloudFoundryReleaseAnalyzer.java
ilayaperumalg: In this case, we support only the single application for CF based release manager. Hence, we return just the app name in use.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/ReleaseAnalyzer.java
ilayaperumalg: I thought about this earlier. My thoughts were since the user wants to force upgrade a specific list of apps the properties correspond to other apps in the `--propertiesFile` should be ignored.

PR #735, opened 2018-10-25
Upgrade to Spring Boot 2.0.x

### LINKED ISSUE #733: Upgrade to Spring Boot 2.x
While we will eventually go with boot 2.1.x, it's easier to first upgrade to 2.0.x as not all deps are not yet ready and most likely we'll take a hit with boot's bean override setting in 2.1.x.

jvalkeal:
- This commit updates and fixes everything so that
  we can base build on Boot2.
- Boot and Cloud deps from parent build 2.0.3.
- Refactor ErrorAttributes from boot
- Use common security which builds atop of framework 5. As essentially
  boot now forces security autoconfig, We can't have @EnableWebSecurity
  in SkipperOAuthSecurityConfiguration anymore.
- Fix repository methods per changes in spring data
- Use flyway version from boot and use its changed settings in boot
- Fixed some issues in oauth tests and still thinking how
  it even worked in 1.x branch.
- Fixes #733

Btw, you now need to use `--spring.config.additional-location` instead of `--spring.config.location` if adding `skipper.yml`. This is a boot change.


spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/statemachine/StateMachineTests.java
```diff
@@ -419,12 +420,24 @@ public void testUpgradeCancelWhileCheckingApps() throws Exception {
 					.step()
 						.sendEvent(message2)
 						.expectStateChanged(2)
-						.expectStateEntered(SkipperStates.UPGRADE_CANCEL,
-								SkipperStates.INITIAL)
+						// for now need to do a trick to wait this later
```
ilayaperumalg: @jvalkeal Could you elaborate on this specific change? I am not sure I understand the reasoning (especially with Boot2 changes).

jvalkeal: There is a this timer loop in a machine which allows to break out from it if we'd do a cancel(which btw is currently not really even used but machine model allows it). There were some changes in a statemachine in 2.x line working little differently what comes for these timers, which then cause these loops to work differently. As I wanted to solely rely on spring-statemachine-test package to execute tests, I now noticed that test step executions are missing some functionality to conditionally wait something to happen. That's why I needed to modify this test to wait it to eventually end up back to INITIAL state. This test case was originally a bit of hack as you get trouble with threads and timers.

But as I said, what this test is testing is not currently used in a skipper as we don't support cancelling an upgrade while used machine theoretically supports it.

ilayaperumalg:
LGTM

ilayaperumalg:
Merging.

PR #793, opened 2019-01-11
Use Spring Cloud Kubernetes to read ConfigMap

### LINKED ISSUE #792: Use Spring Cloud Kubernetes to read ConfigMap
Add Spring Cloud Kubernetes dep to load properties from the application.yaml config block from a k8s ConfigMap.

Relates to https://github.com/spring-cloud/spring-cloud-dataflow/issues/2735

chrisjs:
Resolves #792

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
         <reactor.version>3.1.8.RELEASE</reactor.version>
         <kubernetes-client.version>4.0.4</kubernetes-client.version>
         <spring-cloud-deployer-kubernetes.version>2.0.0.M1</spring-cloud-deployer-kubernetes.version>
+        <spring-cloud-kubernetes.version>0.3.0.RELEASE</spring-cloud-kubernetes.version>
 
         <dockerfile-maven-plugin.version>1.4.9</dockerfile-maven-plugin.version>
         <zeroturnaround.version>1.13</zeroturnaround.version>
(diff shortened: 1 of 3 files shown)


jvalkeal:
If only thing what this brings into skipper is `ConfigReloadAutoConfiguration` I don't find it very useful, unless I'm missing something. I'm expecting whole skipper to blow up if running context is refreshed?

jvalkeal:
Ah it's `BootstrapConfiguration` which gets created automatically and that exception comes from there and I see same with dataflow. I'm not exactly happy about `spring-cloud-kubernetes-config` as its default functionality is way too intrusive.

chrisjs:
its being used to read boot properties from a k8s `ConfigMap`, for example:

https://github.com/spring-cloud/spring-cloud-dataflow/blob/801faa8d9bbcf3c2a6c267660e3ebb4aa6051536/src/kubernetes/skipper/skipper-config-kafka.yaml#L8

while the same is being done in dataflow the main motivator for using it here as well was the account syntax being used -- for example `_[default]_` isn't a valid pattern for a container env var as the regex used for validation is '[-._a-zA-Z][-._a-zA-Z0-9]*'). that would leave us with stuffing things into SAJ which can get a bit out of hand.

jvalkeal:
Merged per c2838cee2757d688ce337c41495331a5a6266ef0

PR #794, opened 2019-01-14
Use MavenProperties from deployer spi

### LINKED ISSUE #788: Rely MavenProperties from deployer spi auto-config
Classes like `EnableSkipperServerConfiguration` should be moved out from server core. `MavenConfigurationProperties` should be removed in favour of one coming from deployer spi.

jvalkeal:
- Import spring-cloud-deployer-dependencies and remove explicit
  versions from poms.
- Remove class MavenConfigurationProperties which is now part of a
  deployer auto-config and it's only needed to autowire and use
  MavenProperties coming out from spring-cloud-deployer-resource-maven.
- Fix all tests to use ResourceLoadingAutoConfiguration
- Fixes #788


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 1b2c0a0

PR #795, opened 2019-01-14
Change default logging config

### LINKED ISSUE #741: startup logs do not print port that skipper is listening on

jvalkeal:
- Fix changed package from org.springframework.boot.context.embedded.tomcat
  to org.springframework.boot.web.embedded.tomcat order to log
  listen port during startup.
- Fixes #741


cppwfs:
LGTM

PR #800, opened 2019-01-24
Flyway compatibility with dataflow

### LINKED ISSUE #799: Flyway compatibility with dataflow
In case same database is used with skipper/dataflow, skipper blows up if it is not started first with an empty schema.

Dataflow will have better model of using flyway to do migrations and we should adapt with same concepts.

jvalkeal:
- This commit tries to provide compatibility with dataflow
  so that they can both co-exist in a same schema and servers
  can be started independently as previously skipper had
  to be started first.
- Move away from sql based scripts by using custom
  SkipperFlywayMigrationStrategy which checks if we're already
  in initial schema and then does a flyway repair programmatically
  which switches from sql scripts to java migrations.
- Remove all sql scripts expect embedded h2 and move that
  under org.springframework.cloud.skipper.server.db.migration.h2
  package as db/migration location is removed from config.
- R__Hibernate_Sequence for mssql now contains hibernate_sequence
  table to sequence fix which is also done programmatically and
  same fix also goes to dataflow(which would be needed if dataflow 2.x
  is started before skipper 1.x to 2.x migration has happened).
- Other than that this pretty much uses same concepts added to dataflow
- No actual tests as server migration combinations are tested
  within db acceptance tests.
- Fixes #799


markpollack:
Tested on PCF upgrading from previous versions of SCDF/Skipper.  worked well.  Clean up some logging statements and squashed.  Merged as fc8e76e009e75e8fe389c1757f32561cc53b902f

PR #805, opened 2019-02-01
Replace "spring.config.location" references with "spring.config.additional-location"

### LINKED ISSUE #771: Revisit usage of spring.config.location
In our docs we document using `--spring.config.location=skipper.yml` to provide additional configuration to Skipper.

However, that is not working - according to Boot docs it will *replace* the yml. In our case the application will most likely fail to start (even with a completely empty yaml file)

Instead it looks like we should be using `--spring.config.additional-location=skipper.yml`

sabbyanandan:
Resolves #771

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -64,7 +64,7 @@ Where you replace `/home/mpollack/projects/` to the path where you cloned `sprin
 Then start the Skipper server
 
 ----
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
LGTM, merging.

PR #809, opened 2019-02-02
Remove connector libraries

### LINKED ISSUE #784: Remove database connector dependency from the classpath
As a developer, I'd like to remove the dependency of spring-cloud-connectors from the classpath; instead, we would switch to Boot's autoconfiguration proper.

markpollack:
- Updated documentation

Tested on CF with mysql service

Fixes #784


ilayaperumalg:
LGTM, merging.

PR #822, opened 2019-02-12
Change package-repositories config from list to map

### LINKED ISSUE #789: Change package-repositories from list to map
Currently we use repos as list.
```
spring:
  cloud:
    skipper:
      server:
        packageRepositories:
          -
            name: experimental
            url: http://skipper-repository.cfapps.io/repository/experimental
            description: Experimental Skipper Repository
            repoOrder: 0
          -
            name: local
            url: http://${spring.cloud.client.hostname}:7577
            local: true
            description: Default local database backed repository
            repoOrder: 1
```
This effectively is hardcoding these into a config as complex list of objects is makes it impossible to override anything and modification needs to re-write everything.

Having these as a map where key is a repo name would suit needs better as then overriding something is accurate and clear.

jvalkeal:
- Change actual configuration class not to use `Repository` entity class
  directly and introduce `PackageRepository` which is then used in
  `RepositoryInitializationService` to wire settings.
- Fix tests and docs.
- With this it's now possible to change values i.e. using
  spring.cloud.skipper.server.package-repositories.local.local=false
- Fixes #789

### DIFF
--- a/spring-cloud-skipper-docs/src/main/asciidoc/three-hour-tour.adoc
+++ b/spring-cloud-skipper-docs/src/main/asciidoc/three-hour-tour.adoc
@@ -795,13 +795,11 @@ spring
     skipper:
       server:
         package-repositories:
-          -
-            name: experimental
+          experimental:
             url: http://skipper-repository.cfapps.io/repository/experimental
             description: Experimental Skipper Repository
             repoOrder: 0
-          -
-            name: local
+          local:
             url: http://${spring.cloud.client.hostname}:7577
             local: true
             description: Default local database backed repository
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
@jvalkeal This means the old properties won't work now. I guess we are ok not supporting the old style when using Skipper 2.x ?

jvalkeal:
Well this is only time to make that breaking change as we can't support that old format.

ilayaperumalg:
LGTM

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 7973ce0

PR #824, opened 2019-02-14
Programmatic setting of profiles

### LINKED ISSUE #796: Auto-detect whether or not the deployed platform is Kubernetes
Prior to starting the Skipper server, we would want to determine if we are Kubernetes. If we are not, we will have to set `spring.cloud.kubernetes.enabled` to "false".

### LINKED ISSUE #807: Update strategy for creating local platform and setting profiles
Data Flow now has a better mechanism for managing the creation of a local platform.  The configuration property `SPRING_CLOUD_SKIPPER_ENABLE_LOCAL_PLATFORM` would no long be needed if we adopt that strategy in Skipper.

Essentially, the local platform deployer is created only if the server is not running on a cloud platfrom and no other deployer is defined.

The profile 'local' will also be set as appropriate, so that profile can be used in logback.xml to configure logging to disk.

Documentation for the CF manifest would also be able to remove setting of the 'cloud' profile.

See https://github.com/spring-cloud/spring-cloud-dataflow/commit/7b0cb2ec4f10c0be2fb6b1fade9e1250def0e157

Relates to spring-cloud/spring-cloud-dataflow#2839.

markpollack:
Set local profile if no cloud profile is detected.
Remove references to enableLocalPlatform
Set spring.cloud.kubernetes.enabled to false if not on kubernetes

Fixes #807
Fixes #796


spring-cloud-skipper-autoconfigure/src/main/java/org/springframework/cloud/skipper/server/autoconfigure/ProfileApplicationListener.java
```diff
@@ -0,0 +1,90 @@
...
+					}
+					environment.addActiveProfile(profileToAdd);
+					addedCloudProfile = true;
+				}
+			}
+		}
+		if (!addedKubernetesProfile) {
+			Map<String, Object> properties = new LinkedHashMap<>();
+			properties.put("spring.cloud.kubernetes.enabled", false);
+			logger.debug("Setting property 'spring.cloud.kubernetes.enabled' to false.");
```
jvalkeal: Could we log on info level and add more logging what this class does? It's such a deep magic that if something goes wrong it'd be good to see what profiles got enabled.

spring-cloud-skipper-autoconfigure/src/main/java/org/springframework/cloud/skipper/server/autoconfigure/ProfileApplicationListener.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author Mark Pollack
+ */
+public class ProfileApplicationListener implements ApplicationListener<ApplicationEnvironmentPreparedEvent>, Ordered {
+
+	private static final Logger logger = LoggerFactory.getLogger(ProfileApplicationListener.class);
+
+	private ConfigurableEnvironment environment;
+
+	@Override
+	public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
```
jvalkeal: How is all this working if `local`, `cloud` or `kubernetes` profile is already set either by user or a platform? Shouldn't we then simply back of?

spring-cloud-skipper-autoconfigure/src/main/java/org/springframework/cloud/skipper/server/autoconfigure/ProfileApplicationListener.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.core.Ordered;
+import org.springframework.core.env.CommandLinePropertySource;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.core.env.MutablePropertySources;
+
+/**
+ * @author Mark Pollack
+ */
+public class ProfileApplicationListener implements ApplicationListener<ApplicationEnvironmentPreparedEvent>, Ordered {
```
jvalkeal: Might also be a good idea to provide a way to disable this functionality. Boot's `org.springframework.boot.autoconfigure.BackgroundPreinitializer` shows how property can be used via system properties. All this happens so early that system property is only one available at this time.

spring-cloud-skipper-autoconfigure/src/main/java/org/springframework/cloud/skipper/server/autoconfigure/ProfileApplicationListener.java
markpollack: :+1:

spring-cloud-skipper-autoconfigure/src/main/java/org/springframework/cloud/skipper/server/autoconfigure/ProfileApplicationListener.java
markpollack: Hmmm... that prob makes sense.  Need to do the same in DataFlow as this code is pretty much duplicated from there.

spring-cloud-skipper-autoconfigure/src/main/java/org/springframework/cloud/skipper/server/autoconfigure/ProfileApplicationListener.java
markpollack: :+1:

jvalkeal:
Think this looks ok, would you mind to do a rebase as there's conflict with `SkipperServerProperties`.

markpollack:
opened new PR, rebased and squashed.

Issue #825, opened 2019-02-15
Exclude UserDetailsServiceAutoConfiguration


ghillert:
When starting Skipper you will see console output such as the following:

```
Using generated security password: f0bfd96f-8629-45a1-8738-bcb417a0d996
```

This is due to `UserDetailsServiceAutoConfiguration` being triggered.

PR #826, opened 2019-02-15
Exclude UserDetailsServiceAutoConfiguration

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-skipper/issues/825


jvalkeal:
Hold on, looking that class and I thought in these days we should not even need to exclude `SecurityAutoConfiguration` as boot should back away when we create our own opinionated config?

jvalkeal:
Created https://github.com/spring-cloud/spring-cloud-common-security-config/issues/36 as I think this should eventually get handled centrally there.

jvalkeal:
Merged per fe5cfaa07c15940b144d0e20d2232b02846df105.

PR #828, opened 2019-02-20
Programmatic setting of profiles

### LINKED ISSUE #796: Auto-detect whether or not the deployed platform is Kubernetes

### LINKED ISSUE #807: Update strategy for creating local platform and setting profiles

markpollack:
Set local profile if no cloud profile is detected.
Remove references to enableLocalPlatform
Set spring.cloud.kubernetes.enabled to false if not on kubernetes
Address review comments

Fixes #807
Fixes #796

Address review comments


jvalkeal:
lgtm

PR #866, opened 2019-04-10
Add deployer properties metadata filtering

jvalkeal:
- Essentially adding 4 new server properties with string array type;
  spring.cloud.skipper.server.deployer-properties.group-excludes,
  spring.cloud.skipper.server.deployer-properties.property-excludes,
  spring.cloud.skipper.server.deployer-properties.group-includes,
  spring.cloud.skipper.server.deployer-properties.property-includes.
- DeployerConfigurationMetadataResolver will use these properties
  to white/black list resolved properties which are prefixed with
  spring.cloud.deployer..
- Add one excluded group `spring.cloud.deployer.kubernetes.fabric8` to
  a server configuration as those keys are expected to come from k8s
  deployer when its metadata polish PR is merged.

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/metadata/DeployerConfigurationMetadataResolver.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/metadata/DeployerConfigurationMetadataResolver.java
@@ -17,23 +17,33 @@ package org.springframework.cloud.skipper.server.deployer.metadata;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 
 import org.springframework.beans.BeansException;
+import org.springframework.boot.configurationmetadata.ConfigurationMetadataGroup;
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataRepository;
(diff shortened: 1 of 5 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/metadata/DeployerConfigurationMetadataResolver.java
```diff
@@ -58,12 +68,47 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+		// 4. pass in group excludes
+		// 5. same logic for properties for includes and excludes
+		// 6. what's left is white/black listed props
+		groups.values().stream()
+			.filter(g -> g.getId().startsWith(KEY_PREFIX))
+			.filter(groupEmptyOrAnyMatch(deployerProperties.getGroupIncludes()))
+			.filter(groupNoneMatch(deployerProperties.getGroupExcludes()))
+			.forEach(g -> {
+				g.getProperties().values().stream()
+					.filter(propertyEmptyOrAnyMatch(deployerProperties.getPropertyIncludes()))
```
ilayaperumalg: Does `includes` mean we should always include the properties? Not sure I understand if this means we send the property with the empty value?

jvalkeal: No, it just means we effectively skip match if include definitions array is empty. It's something we need to do with `anyMatch` as `noneMatch` doesn't need this trick.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 4f44015

PR #878, opened 2019-06-13
Skipper docker image synchronization utility

### LINKED ISSUE #887: Create domain model for Release Log info
Currently, the release logs are retrieved as String. It would be better to have the REST response to be a POJO instead of just a String. Also, we can pass along a map of values that correspond to individual applications.

tzolov:
- Add a wait-for-it.sh script to the top image folder. Later can be used to make the skipper service wait until external dependencies such as DBs are provided.

  Resolves #887

### DIFF
--- a/spring-cloud-skipper-server/Dockerfile
+++ b/spring-cloud-skipper-server/Dockerfile
@@ -3,5 +3,6 @@ FROM springcloud/openjdk:latest
 ARG JAR_FILE
 
 ADD target/${JAR_FILE} spring-cloud-skipper-server.jar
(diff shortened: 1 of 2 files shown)


tzolov:
NOTE: this PR will not be resolved! 
Instead the `wait-for-it.sh` will be added to the parent image (`sprintcloud/openjdk`) as part of: https://github.com/spring-cloud/openjdk-docker/pull/2

tzolov:
What fix as part of Skipper but in the root image as part of: spring-cloud/openjdk-docker#2

PR #879, opened 2019-06-17
Polish build files

### LINKED ISSUE #562: Polish pom files
Should have same format in all poms, not mixing tabs/spaces and different style of indentation.

jvalkeal:
- Tabs vs spaces
- Move all version to main pom.
- Fix wrong version def for spring-cloud-kubernetes
- Remove dep management and versions for commons-lang and snakeyml,
  as we defined older versions while actual version is coming from
  parent build.
- Resulting fatjar is identical for master
- Fixes #562


markpollack:
rebased and merged as jvalkeal-gh562

PR #880, opened 2019-06-18
Update database docs

### LINKED ISSUE #874: Support mysql 8
Doesn't work with current mariadb driver so fix would be to bundle mysql driver. User can then choose a driver without re-building a server.

### LINKED ISSUE #875: Support mariadb 10.3+
Starting from mariadb 10.3 a proper sequence has been added and things we do with a sequence table originating from remains of mysql doesn't work anymore.

Need to fully move to this and think about all migration nonsense this brings.

jvalkeal:
- Essentially copy relevant parts from a dataflow docs as
  db configs are identical.
- Fixes #874
- Fixes #875


markpollack:
edited and merged as 9d95c879424f5797c747062d4a7036fe7a3aa454

PR #881, opened 2019-07-02
Fix changes in a local deployer

jvalkeal:
- We still depend on local deployer metadata and
  need to change tests to match what's expected.
- Essentially skip what might have in a boot's hint
  arrays as it's just noise.

NOTE: this is a fix for local deployer PR https://github.com/spring-cloud/spring-cloud-deployer-local/pull/161 what that gets merged.


ilayaperumalg:
LGTM, merging.

PR #904, opened 2019-08-21
Remove creation of copied default platform

### LINKED ISSUE #901: Revisit single-platform resolution behavior
Notes:
- [ ] If there's only one platform, we would automatically pick that as the default without an explicit entry with "default" key in the map
- [ ] If there are >1 platforms, a user must select a platform to deploy streams (platform won't be automatically inferred in this case)
- [ ] If there's something with "default", we will pick it up as the "default" platform
- [ ] Update the docs/samples/microsite to clarify this

jvalkeal:
- In case user defined a single platform not named as `default`, it
  was duplicated as `default` resulting two platforms. This commit
  simply removes this duplication.
- Internally we don't do any dance around platform names so that we'd
  silently switch to `default` if only one non `default` exists. This
  is because similar check is allready done on a dataflow side and this
  switch happens there. In skipper we use these platforms via repository
  and it would have added a complex changes to switch these names in
  multiple places, essentially doing platform switch in both skipper
  and dataflow. It's just better to fail on a skipper side if
  non-existing platform is used as that keeps logic simpler.
- Fixes #901

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/DeployerInitializationService.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/service/DeployerInitializationService.java
@@ -15,7 +15,6 @@
  */
 package org.springframework.cloud.skipper.server.service;
 
-import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

Issue #910, opened 2019-10-18
Implement scale() operation for SkipperClient


jvalkeal:
Apart from a core changes for scaling, order to use scaling from scdf SkipperClient needs to implement same features.

jvalkeal:
This is now handled in PR #911

PR #911, opened 2019-10-18
Implement scaling for request flows

### LINKED ISSUE #909: Implement scale() operation
Depending on how new scale API is added to deployers, it has to be added to current release upgrade flow in a statemachine and then exposed to a dataflow server via rest api.

Theoretically if applicable we could just use count property meaning from a scdf point of view it would just update stream with `deployer.*.count=2` setting and skipper would then call appropriate deployer api's. Currently if you update stream with new count, nothing happens as there's no differences in an actual apps.

jvalkeal:
- Basically adding a new state flow `SCALE` via statemachine
  which as a simple passthrough dispatches to deployer scale method.
- 2 new rest methods for single app and multi app scaling.
- Needed SkipperClient changes are left for #910
- Fixes #909

For now, you can test this i.e. with existing `ticktock` with this rest call:
```
POST http://localhost:7577/api/release/scale/ticktock
Content-Type: application/json

{
    "counts": {
        "time": "2",
        "log": "2"
    }
}
```

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
@@ -152,13 +152,13 @@ public class ReleaseController {
 		return this.releaseResourceAssembler.toModel(release);
 	}
 
-	@RequestMapping(path = "/scale/{name}/{appName}", method = RequestMethod.POST)
+	@RequestMapping(path = "/scale/{name}/{appName}/{count}", method = RequestMethod.POST)
 	@ResponseStatus(HttpStatus.CREATED)
 	public EntityModel<Release> scale(@PathVariable("name") String name, @PathVariable("appName") String appName,
-			int desiredCount) {
+			@PathVariable("count") int count) {
 		ScaleRequest scaleRequest = new ScaleRequest();
 		Map<String, Integer> counts = new HashMap<>();
-		counts.put(appName, desiredCount);
+		counts.put(appName, count);
 		scaleRequest.setCounts(counts);
 		Release release = this.skipperStateMachineService.scaleRelease(name, scaleRequest);
 		return this.releaseResourceAssembler.toModel(release);
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
@@ -15,9 +15,7 @@
  */
 package org.springframework.cloud.skipper.server.controller;
 
-import java.util.HashMap;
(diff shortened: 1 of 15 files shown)


spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/ScaleRequest.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.springframework.cloud.skipper.domain;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Janne Valkealahti
+ */
+public class ScaleRequest {
```
tzolov: Deployer's scale domain time take deployerId, count and optional Map<Sting, String>. 
To represent this on the Skipper side the `ScaleRequest` class, perhaps can have a structure something like this: 
```java
public class ScaleRequest {
   private String name; // appName (not deployerId)
   private String count; // instance count
   private Map<String, String> properties; // optional
   // ... Getters/Setters ...
}
```
And then to represent multiple app scales we can use `List<ScaleRequest>`. Above should match to JSON like:

```json
[ 
 {"name": "time", "count": "2", "properties": {...}},
 {"name": "log", "count": "3", "properties": {...}},
 ...
]
```

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
```diff
@@ -142,6 +143,13 @@ public ReleaseControllerLinksResource resourceLinks() {
 		return this.manifestResourceAssembler.toModel(this.releaseService.manifest(name, version));
 	}
 
+	@RequestMapping(path = "/scale/{name}", method = RequestMethod.POST)
+	@ResponseStatus(HttpStatus.CREATED)
+	public EntityModel<Release> scale(@PathVariable("name") String name, @RequestBody ScaleRequest scaleRequest) {
```
ilayaperumalg: Do you think it would be better if we just have the `ScaleRequest` as the Request body and remove the explicit name from the path (given the ScaleRequest already has the name in it)?

spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/ScaleRequest.java
jvalkeal: done

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/controller/ReleaseController.java
jvalkeal: Not sure, I followed same conventions where release name is always coming from a request path.

ilayaperumalg:
Squashed, rebased and merged as 2fb32516b15394cbc9a19559426a3d575b0045ee

PR #917, opened 2019-10-25
Remove hateoas from a SkipperClient

### LINKED ISSUE #913: Shell broken for traverson link following
This looks to be one migration issue with spring-hateoas 1.x. Currently in master we get nothing for commands which rely on traverson:
```
skipper:>platform list
╔════╤════╤═══════════╗
║Name│Type│Description║
╚════╧════╧═══════════╝

skipper:>repo list
╔════╤═══╤═════╤═════╗
║Name│URL│Local│Order║
╚════╧═══╧═════╧═════╝
```

jvalkeal:
NOTE: There is a companion PR on a dataflow side for this https://github.com/spring-cloud/spring-cloud-dataflow/pull/3579

- As hateoas configuration currently doesn't work outside
  of a servlet environment(as shell is running), fixing
  this issue by removing hateoas.
- Basically removing hateoas wrappers and changing api
  signatures to contain plain java classes.
- As this is basically SkipperClient api polish, also
  remove add/remove methods for repository which were
  never used or supported anyway.
- Fixes #913

### DIFF
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
@@ -16,14 +16,14 @@
 package org.springframework.cloud.skipper.client;
 
 import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.nio.charset.Charset;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
```diff
@@ -317,8 +320,8 @@ public Release rollback(String releaseName, int releaseVersion) {
 
 	@Override
 	public List<Release> list(String releaseNameLike) {
-		ParameterizedTypeReference<CollectionModel<Release>> typeReference = new ParameterizedTypeReference<CollectionModel<Release>>() {
-		};
+		ParameterizedTypeReference<HateoasResponseWrapper<ReleasesResponseWrapper>> typeReference =
```
ilayaperumalg: How is this different from using just the  `Release` domain object instead of explicit wrapper? I have the same question for other places where the ResponseWrappers are used.

jvalkeal: Server side is still sending hateoas json so wanted to completely remove hateoas classes from client. This is still needed for jackson.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as a60747ce

PR #941, opened 2020-02-23
Add new api to request multiple stream statuses

### LINKED ISSUE #940: Add a new controller to adapt to reactive AppStatus changes
This is a related part for spring-cloud/spring-cloud-deployer#336 to implement reactive api's.

Also relates to spring-cloud/spring-cloud-skipper#939 for perf enhancements. 

~Part for performance is also to cache some deployer requests not to cause too much load for things which are not expected to change too frequently to suit multiple users.~

jvalkeal:
- ReleaseController now has a new method /statuses
  which returns a map of statuses wrapped in a mono
  and subscription is then handled by framework itself.
- ReleaseManager has a new method to request status reactively
  with a Release.
- SkipperClient has a new statuses method for dataflow to request
  info for multiple releases at a same time.
- DefaultReleaseManager now tries to use reactive methods for statuses
  and doesn't a bit of caching so that we don't ddos our own server.
- ReleaseStatUpdateService now also uses reactive methods so that update
  is faster by not doing things sequentially. There's also a little fix
  to do update immediately when server starts(via initial polling flag).
- Fixes #940

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/DefaultReleaseManager.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/DefaultReleaseManager.java
@@ -32,7 +32,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
-import reactor.util.function.Tuple2;
 
 import org.springframework.cloud.deployer.spi.app.AppDeployer;
 import org.springframework.cloud.deployer.spi.app.AppInstanceStatus;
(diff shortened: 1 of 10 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/deployer/DefaultReleaseManager.java
```diff
@@ -238,6 +257,70 @@ private void updateCountProperty(Map<String, Object> model, String appsCount) {
...
+			List<String> deploymentIds = appDeployerData.getDeploymentIds();
+			logger.info("Getting status for {} using deploymentIds {}", release,
+					StringUtils.collectionToCommaDelimitedString(deploymentIds));
+
+			if (!deploymentIds.isEmpty()) {
+				Map<String, String> appNameDeploymentIdMap = appDeployerData.getDeploymentDataAsMap();
+				return Flux.fromIterable(appNameDeploymentIdMap.entrySet())
+					.flatMap(nameDeploymentId -> {
+						String deploymentId = nameDeploymentId.getValue();
+						return appDeployer.statusReactive(deploymentId);
```
ilayaperumalg: This call still is made to retrieve ApplicationDetail while we can limit this to fetch only the summary if that is all that we need.  We can split the statusReactive(Release release) into two parts 1)with ApplicationDetail 2) without ApplicationDetail.

jvalkeal:
Merged with d4a4c444d9453a50893719e8d51dc675b15ec5c6

PR #949, opened 2020-03-31
Update spring-doc-resources and code highlight

Buzzardo:
Update the spring-doc-resources version to pick up Damien's latest changes
to the appearance of the docs.


Buzzardo:
Why would we not want to upgrade to the latest version and change the highlighter?

PR #958, opened 2020-06-18
Avoid NPE in AppDeployerData

### LINKED ISSUE #957: Handle AppDeployerData when it is not available for a release
If a release doesn't have an `AppDeployerData` associated with it, we need to handle the operations related to `AppDeployerData` appropriately. 

When the `AppDeployerData` domain was designed we always enforced that it should be available for the release but, it turns out that in some cases, there could be missing AppDeployerData. For instance, if the config server isn't available to retrieve app metadata etc.,

ilayaperumalg:
- Though the `AppDeployerData` is expected to be non-null, we have seen some scenarios where NPEs are thrown when retrieving the AppDeployerData.
   Hence, added null check at all the places where `AppDeployerData` is used.

Resolves #957


jvalkeal:
lgtm

PR #967, opened 2020-08-03
Fix changed cf java client

jvalkeal:
- buildpack is changed as buildpacks in a manifest
  in a new cf java client.
- You can still set single buildpack but can read
  back only multiple ones.
- This commit needs cf deployer which brings in cf 4.9.x


ilayaperumalg:
LGTM, merging.

PR #973, opened 2020-08-28
Update doc generation

Buzzardo:
Update doc generation to the latest versions of spring-doc-resources
and the Asciidoctor generator. Also removed tocify, which was
causing a duplicate ToC. Also removed the last bit of XML leftover
from using Docbook.


jvalkeal:
Merged per a9b3f187e21cf4217e0be949189f9258c0ce920b

PR #1025, opened 2022-01-19
Add ActuatorService and ReleaseController endpoints

### LINKED ISSUE #1024: Expose ActuatorOperations in API
See https://github.com/spring-cloud/spring-cloud-dataflow/issues/4758 
Add `/actuator` endpoints to the skipper API

dturanski:
resolves #1024 

The `/actuator`endpoint goes under the `/release` resource.  Following the precedent for `api/release/logs`, the functionality is similarly out of the release domain, but gets info provided by deployed app instances whose metadata are exposed via a release.  Exposed via `ReleaseController`, this adds an `ActuatorService` which also depends on `ReleaseService`, and config for the ActuatorService.  I have manually verified end-to-end functionality against the local deployer, with `time | log` deployed with 2 log instances.   

Depends on https://github.com/spring-cloud/spring-cloud-deployer-cloudfoundry/pull/385, https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/478, and https://github.com/spring-cloud/spring-cloud-deployer-local/pull/212

### DIFF
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 the original author or authors.
+ * Copyright 2017-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,7 @@ import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.deployer.spi.app.DeploymentState;
 import org.springframework.cloud.skipper.domain.AboutResource;
+import org.springframework.cloud.skipper.domain.ActuatorPostRequest;
 import org.springframework.cloud.skipper.domain.CancelRequest;
 import org.springframework.cloud.skipper.domain.CancelResponse;
 import org.springframework.cloud.skipper.domain.Deployer;
@@ -62,6 +63,7 @@ import org.springframework.web.util.UriComponentsBuilder;
  * @author Mark Pollack
  * @author Ilayaperumal Gopinathan
  * @author Janne Valkealahti
+ * @author David Turanski
  */
 public class DefaultSkipperClient implements SkipperClient {
 
@@ -235,6 +237,43 @@ public class DefaultSkipperClient implements SkipperClient {
 						httpEntity,
 						typeReference,
 						uriVariables);
+		return resourceResponseEntity.getBody();
+	}
+
+	@Override
+	public String getFromActuator(String releaseName, String appName, String appId, String endpoint) {
+
+		Map<String, Object> uriVariables = new HashMap<>();
+		uriVariables.put("releaseName", releaseName);
+		uriVariables.put("appName", appName);
+		uriVariables.put("appId", appId);
+
+		ResponseEntity<String> resourceResponseEntity =
+				restTemplate.exchange(
+						baseUri + "/release/actuator/{releaseName}/{appName}/{appId}?endpoint=" + endpoint,
+						HttpMethod.GET,
+						null,
+						String.class,
+						uriVariables);
+
+		return resourceResponseEntity.getBody();
+	}
+
+	@Override
+	public Object postToActuator(String releaseName, String appName, String appId, ActuatorPostRequest request) {
+		Map<String, Object> uriVariables = new HashMap<>();
+		uriVariables.put("releaseName", releaseName);
+		uriVariables.put("appName", appName);
+		uriVariables.put("appId", appId);
+
+		HttpEntity<ActuatorPostRequest> httpEntity = new HttpEntity<>(request);
+		ResponseEntity<String> resourceResponseEntity =
+				restTemplate.exchange(baseUri + "/release/actuator/{releaseName}/{appName}/{appId}",
+						HttpMethod.POST,
+						httpEntity,
+						String.class,
+						uriVariables);
+
 		return resourceResponseEntity.getBody();
 	}
 
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
@@ -267,11 +267,11 @@ public class DefaultSkipperClient implements SkipperClient {
 		uriVariables.put("appId", appId);
 
 		HttpEntity<ActuatorPostRequest> httpEntity = new HttpEntity<>(request);
-		ResponseEntity<String> resourceResponseEntity =
+		ResponseEntity<?> resourceResponseEntity =
 				restTemplate.exchange(baseUri + "/release/actuator/{releaseName}/{appName}/{appId}",
 						HttpMethod.POST,
 						httpEntity,
-						String.class,
+						Object.class,
 						uriVariables);
 
 		return resourceResponseEntity.getBody();
--- a/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/SkipperClient.java
+++ b/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/SkipperClient.java
@@ -242,19 +242,21 @@ public interface SkipperClient {
 	Release scale(String releaseName, ScaleRequest scaleRequest);
 
 	/**
+	 * Access an actuator resource for a deployed app instance.
 	 *
 	 * @param releaseName the release name
 	 * @param appName the application name
 	 * @param appId the deployer assigned guid of the app instance
 	 * @param endpoint the relative actuator path, e.g., {@code /info}
-	 * @return the contents as JSON text.
+	 * @return the contents as JSON text
 	 */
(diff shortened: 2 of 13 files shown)


spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
```diff
@@ -238,6 +240,43 @@ public Release scale(String releaseName, ScaleRequest scaleRequest) {
...
+	public String getFromActuator(String releaseName, String appName, String appId, String endpoint) {
+
+		Map<String, Object> uriVariables = new HashMap<>();
+		uriVariables.put("releaseName", releaseName);
+		uriVariables.put("appName", appName);
+		uriVariables.put("appId", appId);
+
+		ResponseEntity<String> resourceResponseEntity =
+				restTemplate.exchange(
+						baseUri + "/release/actuator/{releaseName}/{appName}/{appId}?endpoint=" + endpoint,
```
onobc, line 253: **[UNRELATED]**: I would expect the `releaseName` to come after `release` rather than `actuator` in the path. Such as `/release/{releaseName}/(actuator|scale|etc)`.

I see the other apis do the same. Obviously not asking for a change here, just wanting to understand if it was intentional etc..

spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
```diff
@@ -238,6 +240,43 @@ public Release scale(String releaseName, ScaleRequest scaleRequest) {
...
+
+		HttpEntity<ActuatorPostRequest> httpEntity = new HttpEntity<>(request);
+		ResponseEntity<String> resourceResponseEntity =
+				restTemplate.exchange(baseUri + "/release/actuator/{releaseName}/{appName}/{appId}",
+						HttpMethod.POST,
+						httpEntity,
+						String.class,
+						uriVariables);
+
+		return resourceResponseEntity.getBody();
```
onobc, line 277: I believe this is always returning a `String` but the API return is typed to `Object`. The `get` counterpart also returns `String`.

spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/SkipperClient.java
```diff
@@ -241,4 +240,23 @@ static SkipperClient create(String baseUrl) {
 	 * @return the status info of a release
 	 */
 	Release scale(String releaseName, ScaleRequest scaleRequest);
+
+	/**
+	 *
```
onobc, line 245: Do you mind adding a high-level comment about what this is doing?

spring-cloud-skipper-client/src/test/java/org/springframework/cloud/skipper/client/DefaultSkipperClientTests.java
```diff
@@ -210,4 +213,38 @@ public void testReleaseList() {
...
+				Collections.singletonMap("state","STOPPED"));
+		MockRestServiceServer mockServer = MockRestServiceServer.bindTo(restTemplate).build();
+		mockServer
+				.expect(requestTo("/release/actuator/tiktok/log/tiktok-log-0"))
+				.andExpect(content().json(
+						"{\"endpoint\":\"/bindings/input\",\"body\":{\"state\":\"STOPPED\"}}"))
+				.andRespond(withSuccess());
+
+		Object response = skipperClient.postToActuator(
+				"tiktok","log", "tiktok-log-0", actuatorPostRequest);
```
onobc, line 248: I think you forgot the asserts ;)

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerPlatformConfiguration.java
```diff
@@ -46,36 +52,59 @@ public DeployerInitializationService deployerInitializationService(DeployerRepos
...
 	@Profile("local")
-	public Platform localDeployers(LocalPlatformProperties localPlatformProperties) {
-		List<Deployer> deployers = new ArrayList<>();
-		Map<String, LocalDeployerProperties> localDeployerPropertiesMap = localPlatformProperties.getAccounts();
-		if (localDeployerPropertiesMap.isEmpty()) {
-			localDeployerPropertiesMap.put("default", new LocalDeployerProperties());
+	@Configuration
+	static class LocalPlatformConfiguration {
+		@Bean
+		@Primary
```
onobc, line 59: Why is `@Primary` needed now?

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerPlatformConfiguration.java
```diff
@@ -46,36 +52,59 @@ public DeployerInitializationService deployerInitializationService(DeployerRepos
 		return new DeployerInitializationService(deployerRepository, platforms, resolver);
 	}
 
-	@Bean
 	@Profile("local")
```
onobc, line 55: **[UNRELATED]** Is skipper only configured for local profile?

spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/controller/ReleaseControllerTests.java
```diff
@@ -306,6 +316,33 @@ public DeploymentState getState() {
...
+
+		mockMvc
+				.perform(get("/api/release/actuator/myTicker/myTicker.log-v1/myTicker.log-v1-0?endpoint=info"))
+				.andExpect(status().isOk()).andReturn();
+
+		verify(actuatorService, times(1))
+			.getFromActuator("myTicker", "myTicker.log-v1", "myTicker.log-v1-0","info");
+
+
+		reset(actuatorService);
```
onobc, line 332: Any reason to not split these into separate tests?

spring-cloud-skipper-client/src/test/java/org/springframework/cloud/skipper/client/DefaultSkipperClientTests.java
dturanski: the `expects` are actually asserts on the request. I don't care about the response.

spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
dturanski: Get returns a json string, otherwise, the user needs to pass the response type which would be a PITA.  Similar reasoning for post, but user isn't always concerned about the response body, just the http status.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerPlatformConfiguration.java
dturanski: By default, its local. The `kubernetes` or `cloud` profiles are set at runtime if deployed to other platforms.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerPlatformConfiguration.java
dturanski: Because There are multiple `Platform` instances configured in the context always. The ActuatorService for the `local` profile  gets the deployer from this one.

spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/controller/ReleaseControllerTests.java
dturanski: I merged them because `install` takes some time.

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/config/SkipperServerPlatformConfiguration.java
onobc: Might be worth adding this as a comment. Anytime I see `@Primary` it makes me "pump my brakes" and ask "why is this here?".

spring-cloud-skipper-client/src/test/java/org/springframework/cloud/skipper/client/DefaultSkipperClientTests.java
onobc: Ahh yeh, got it. Do you mind adding this as a 1 liner comment?

spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/DefaultSkipperClient.java
onobc: Yep, I was pretty sure that the main reasoning for dropping the Type support was "PITA". :) 

Might we worth calling out something about these return types in the javadocs (if you did not already).

onobc:
@dturanski changes look good - thank you for that.

dturanski:
closing - replaced by PR #1032

PR #1035, opened 2022-03-14
spring-cloud-skipper-1034 Tests fail under Windows

corneil:
Remove `CR (\r)`  for comparison to files.

### DIFF
--- a/spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/service/ArgumentSanitizerTests.java
+++ b/spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/service/ArgumentSanitizerTests.java
@@ -36,7 +36,7 @@ public class ArgumentSanitizerTests {
 				TestResourceUtils.qualifiedResource(getClass(), "nopassword.yaml").getInputStream(),
 				Charset.defaultCharset());
 		String result = ArgumentSanitizer.sanitizeYml(initialYaml);
-		assertThat(result).isEqualTo(initialYaml);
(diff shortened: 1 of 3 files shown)


onobc:
Hi @corneil , we typically follow [these guidelines](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) for commit messages.

corneil:
@onobc Is the new commit comment fine?

onobc:
> @onobc Is the new commit comment fine?

Yep those messages look good. We can squash/merge into a single commit before merge which I envision would look something like:

```java
Remove CR (\r) for comparison to files.

Fixes #1034
```

spring-cloud-skipper-server-core/src/test/java/org/springframework/cloud/skipper/server/service/ArgumentSanitizerTests.java
```diff
@@ -36,7 +36,7 @@ public void testNoChange() throws Exception {
 				TestResourceUtils.qualifiedResource(getClass(), "nopassword.yaml").getInputStream(),
 				Charset.defaultCharset());
 		String result = ArgumentSanitizer.sanitizeYml(initialYaml);
-		assertThat(result).isEqualTo(initialYaml);
+		assertThat(result.replace("\r", "")).isEqualTo(initialYaml.replace("\r", ""));
```
onobc, line 39: In general, this makes me worry that we will play wak-a-mole throughout other code finding places like this. 

1. Is there something we can change in the underlying saved files to avoid having to special case the code? 

2. If we can't fix in the file, wdyt about moving this to a utility method? (`assertEqualsIgnoringXYZ()`)

PR #1037, opened 2022-03-17
Change CRLF to LF in tests when comparing files.

### LINKED ISSUE #1034: Tests fail under Windows
Some tests fail when running on Windows while the content seems visibly similar.
One of the text blocks has CRLF and the other has LF only.

corneil:
Change CRLF (\r\n) to LF (\n) in tests when comparing files.

Fixes #1034


jvalkeal:
lgtm

PR #1041, opened 2022-04-05
Remove deprecated dependency spring-security-oauth2

corneil:
Remove `org.springframework.security.oauth:spring-security-oauth2` and tests that require the deprecated project.


onobc (commented):
@corneil changes look straight-forward. My one question w/ this removal is where will be verifying OAuth2 behavior now? ATs? Do we need a follow on ticket(s) for this work if so?

corneil:
> @corneil changes look straight-forward. My one question w/ this removal is where will be verifying OAuth2 behavior now? ATs? Do we need a follow on ticket(s) for this work if so?

Had a chat with @jvalkeal, these test are invalid because the security implementation has actually been migrated away from the deprecated library. The integration test in Dataflow does complete test along with running Keycloak.

onobc (approved):

corneil:
Resolved #1042 and [https://github.com/spring-cloud/spring-cloud-common-security-config/issues/93](https://github.com/spring-cloud/spring-cloud-common-security-config/issues/93)

PR #1059, opened 2022-08-31
Add support for MySQL 5/8 via MariaDB driver

onobc:
* Use FlywayVendorReplacingEnvironmentPostProcessor to replace {vendor} in spring.flyways.locations
* Add back in the MySQL schema
* Use vendor (mysql) not driver in flyway configuration customizer

See https://github.com/spring-cloud/spring-cloud-dataflow/issues/4887

> ⚠️ ⚠️ **depends on** https://github.com/spring-cloud/spring-cloud-dataflow-common/pull/22

### DIFF
--- a/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/db/migration/SkipperFlywayConfigurationCustomizer.java
+++ b/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/db/migration/SkipperFlywayConfigurationCustomizer.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.springframework.cloud.skipper.server.db.migration;
 
 import javax.sql.DataSource;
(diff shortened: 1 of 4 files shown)


spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/db/migration/SkipperFlywayConfigurationCustomizer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2022 the original author or authors.
+ * Copyright 2019-2021 the original author or authors.
```
onobc, line 2: Wow - time travel!

spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/db/migration/SkipperFlywayConfigurationCustomizer.java
```diff
@@ -69,11 +84,25 @@ else if (databaseDriver == DatabaseDriver.DB2) {
...
 			return DatabaseDriver.fromJdbcUrl(url);
 		}
 		catch (MetaDataAccessException ex) {
 			throw new IllegalStateException(ex);
 		}
 	}
+
+	private boolean vendorIsMySql(DataSource dataSource) {
+		try {
+			return Optional.ofNullable(ReflectionUtils.findMethod(dataSource.getClass(), "getJdbcUrl"))
```
onobc, line 97: This impl is inconsistent w/ the same one in SCDF (I will make consistent before merging either).

I am actually going to move this logic into a DatabaseMetadata class that is in common flyway and both of these can use the same code (like I did for the flyway EPP).

spring-cloud-skipper-server-core/src/main/resources/application.yml
```diff
@@ -192,7 +192,7 @@ spring:
     baselineOnMigrate: true
     locations:
       - classpath:org/springframework/cloud/skipper/server/db/migration/{vendor}
-    check-location: false
```
onobc, line 195: This is deprecated (named something else now) but the default is false here and on the new property - I yanked it.

PR #1061, opened 2022-09-08
Update database section of ref guide

onobc:
This handles the Dataflow ref guide updates - Skipper to follow.

### Point1
I have added a "Add custom JDBC driver" section only to remain consistent w/ the SCDF docs. I believe this will not be the final suggested mechanism for users to extend the classpath as @corneil has another way that is less intrusive. We should update the ref guides accordingly (both Dataflow and Skipper - creating an issue so as to not drop the ball on this detail).

### Point2
I have attached generated PDF for your viewing pleasure while reviewing in case you do not have the IDEA asciidoc plugin etc.. (the PDF will NOT be checked in).
- [installation.pdf](https://github.com/spring-cloud/spring-cloud-skipper/files/9529618/installation.pdf)

Closes #5079

### DIFF
--- a/spring-cloud-skipper-docs/src/main/asciidoc/configuration-database.adoc
+++ b/spring-cloud-skipper-docs/src/main/asciidoc/configuration-database.adoc
@@ -1,0 +1,65 @@
+
+[[configuration-database-overview]]
+A relational database is used to store stream and task definitions as well as the state of executed tasks.
+Spring Cloud Skipper provides schemas for *MariaDB*, *MySQL*, *Oracle*, *PostgreSQL*, *Db2*, *SQL Server*, and *H2*. The schema is automatically created when the server starts.
+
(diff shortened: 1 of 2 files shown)


spring-cloud-skipper-docs/src/main/asciidoc/configuration-database.adoc
```diff
@@ -0,0 +1,65 @@
+
+[[configuration-database-overview]]
+A relational database is used to store stream and task definitions as well as the state of executed tasks.
```
cppwfs, line 3: Do we want to mention tasks in skipper?   It has no reference to them.   If we run in Stream-Only mode those tables are not used.

onobc: Yep, good catch.... it was like that [before](https://github.com/spring-cloud/spring-cloud-skipper/pull/1061/files#diff-a7435791a7ff08ddcbdbb5c68fb777d69c91e8600d8340f02900eb5193e92a7bL223) but it should only talk about streams here for sure. I will adjust.

